PR #5, opened 2023-02-27
feat: add kakfa producer

kruskall:


kafka/producer.go
```diff
@@ -17,4 +17,198 @@
...
+// Close stops the producer
+func (p *Producer) Close() error {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.client.Close()
+	return nil
+}
+
+// ProjectIDKey represents the key for the project ID set in the context.
+type ProjectIDKey struct{}
```
kruskall, line 145: **issue**: this is a dirty hack to pass the project id for the PoC. Did we decide on a standard way ?

marclop: there's a `queuecontext` package in this repository which has helpers to set and extract the project ID to/from the context

kafka/producer.go
```diff
@@ -17,4 +17,198 @@
...
+	}
+	// Issue a metadata refresh request on construction, so the broker list is
+	// populated.
+	client.ForceMetadataRefresh()
+
+	cfg.txTopic = fmt.Sprintf("%s.transaction", cfg.TopicPrefix)
+	cfg.spanTopic = fmt.Sprintf("%s.span", cfg.TopicPrefix)
+	cfg.metricTopic = fmt.Sprintf("%s.metric", cfg.TopicPrefix)
+	cfg.errorTopic = fmt.Sprintf("%s.error", cfg.TopicPrefix)
+	cfg.logTopic = fmt.Sprintf("%s.log", cfg.TopicPrefix)
```
marclop, line 128: I'm not sure how I feel about having this here, I think this may not belong in this repository. Instead, I think it would be best to have a list of record mutators that can be passed as configuration. This would provide a cleaner separation between business and implementation logic:

```go
type RecordMutator func(model.APMEvent, *kgo.Record) error

// ProducerConfig holds configuration for publishing events to Kafka.
type ProducerConfig struct {
	// ...
	Mutators []RecordMutator
	// ...
}
```

Not all producers will always produce using the same topics, and we may want to change that independently too in the future (and I believe even now).

kafka/producer.go
```diff
@@ -17,4 +17,198 @@
...
+	defer p.mu.Unlock()
+	p.client.Close()
+	return nil
+}
+
+// ProjectIDKey represents the key for the project ID set in the context.
+type ProjectIDKey struct{}
+
+// ProcessBatch publishes the events in batch to the specified Kafka topic.
+func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch, wg *sync.WaitGroup) error {
```
marclop, line 148: This method needs to implement `model.BatchProcessor`. There can't be a `sync.WaitGroup` as a third argument.
```suggestion
func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch) error {
```

kafka/producer.go
```diff
@@ -17,4 +17,198 @@
...
+
+	// Key, if non-nil, is used to set the key for an event message.
+	//
+	// Producer uses murmur2 partitioning: consistent hashing over the
+	// message key, if it is non-nil; random partition assignment if the
+	// message has no key.
+	//
+	// If Key is nil, the message key will be set to the event's trace ID
+	// if there is one, and nil otherwise.
+	Key func(model.APMEvent) []byte
```
marclop, line 63: If we're to implement the record mutators, it wouldn't be necessary to accept this argument and a Key mutator could be passed in instead.

kafka/producer.go
```diff
@@ -17,4 +17,198 @@
...
+
+	// TODO(axw) Kafka credentials
+
+	// Kafka topics to publish to.
+	txTopic     string
+	spanTopic   string
+	metricTopic string
+	errorTopic  string
+	logTopic    string
```
marclop, line 72: We could also get rid of these fields in this repository.

marclop (commented):
Most of the code looks good, but I think we need to provide a clearer separation between business and adapter logic. Have a look at my proposal and let me know what you think.

kafka/producer.go
axw: As mentioned on Slack, I think we should be passing more generic "metadata" (e.g. `map[string]string`) in/out of context, rather than a project ID. Similar to how gRPC metadata is added to context. That'll give us a bit more flexibility, and separate out some more business logic.

Would it be reasonable to remove `queue.NewConsumer` and `queue.NewProducer`, and require the caller to call the appropriate package method? Then context functions could be directly in the `queue` package without creating a cycle.

marclop: opened https://github.com/elastic/apm-queue/pull/9 which implements the metadata passing. This should allow use to remove this type and rely on the metadata.

kruskall: Update to use the `queuecontext` package with the new methods. Thanks everyone :bow:

kafka/producer.go
kruskall: Good point! Updated :+1:

kafka/producer.go
kruskall: Ah, thanks for this! Changed!

kafka/producer.go
kruskall: Removed!

kafka/producer.go
kruskall: Removed!

kafka/producer.go
```diff
@@ -17,4 +17,164 @@
...
+	// useful since it shows up in Kafka metrics and logs.
+	Version string
+
+	// Logger is used for logging producer errors.
+	Logger *zap.Logger
+
+	// Sync can be used to indicate whether production should be synchronous.
+	Sync bool
+
+	Mutators []RecordMutator
```
marclop, line 56: Can you add a GoDoc description for this field?

kafka/producer.go
```diff
@@ -17,4 +17,164 @@
...
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	p.client.Close()
+	return nil
+}
+
+// ProcessBatch publishes the events in batch to the specified Kafka topic.
+func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch) error {
+	// Take a read lock to prevent Close from closing the input channel
+	// while we're attempting to send to it.
```
marclop, line 122:
```suggestion
	// Take a read lock to prevent Close from closing the client
	// while we're attempting to produce records.
```

kafka/producer.go
```diff
@@ -17,4 +17,164 @@
...
+		}
+
+		encoded, err := event.MarshalJSON()
+		if err != nil {
+			return err
+		}
+
+		record.Value = encoded
+
+		p.client.Produce(ctx, record, func(msg *kgo.Record, err error) {
```
marclop, line 157: I think the code is a bit more shorter and readable like this:
```suggestion
		record := kgo.Record{Headers: headers}
		for _, rm := range p.cfg.Mutators {
			rm(event, &record)
		}
		encoded, err := event.MarshalJSON()
		if err != nil {
			return err
		}
		record.Value = encoded
		p.client.Produce(ctx, &record, func(msg *kgo.Record, err error) {
```

kafka/producer.go
```diff
@@ -17,4 +17,164 @@
...
+
+		if len(headers) != 0 {
+			record.Headers = headers
+		}
+
+		for _, rm := range p.cfg.Mutators {
+			rm(event, record)
+		}
+
+		encoded, err := event.MarshalJSON()
```
marclop, line 150: Can you introduce an `Encoder` interface in this package and use the `codec.JSON` as the producer's encoder?

```go
// Encoder encodes a model.APMEvent to a []byte
type Encoder interface {
	// Encode accepts a model.APMEvent and returns the encoded representation.
	Encode(model.APMEvent) ([]byte, error)
}
```

marclop, line 150: Can you introduce an `Encoder` interface in this package and use the `codec.JSON` as the producer's encoder?

```go
// Encoder encodes a model.APMEvent to a []byte
type Encoder interface {
	// Encode accepts a model.APMEvent and returns the encoded representation.
	Encode(model.APMEvent) ([]byte, error)
}
```

Then this would call the decoder instead of making assumptions on the encoding:
```suggestion
		encoded, err := p.cfg.Encoder.Encode(event)
```

kafka/producer.go
```diff
@@ -17,4 +17,164 @@
...
+	"go.uber.org/zap"
+
+	"github.com/twmb/franz-go/pkg/kgo"
+	"github.com/twmb/franz-go/plugin/kzap"
+
+	"github.com/elastic/apm-data/model"
+	"github.com/elastic/apm-queue/queuecontext"
+)
+
+type RecordMutator func(model.APMEvent, *kgo.Record) error
```
marclop, line 35: Could you add a GoDoc comment here describing the purpose of an individual RecordMutator?

marclop (commented):
Looking good, just a couple of things to be done and it's good to be merged

kafka/producer.go
kruskall: Good catch! Updated!

kafka/producer.go
kruskall: Changed! :+1:

kafka/producer.go
kruskall: Added!

fwiw I think we should move the encoder/decoder interface to the codec package to avoid duplicating it across the producers.

kruskall: Thanks for this! I missed the codec changes :facepalm:

kafka/producer.go
kruskall: Added!

kafka/producer.go
```diff
@@ -17,4 +17,170 @@
...
+	Logger *zap.Logger
+
+	// Encoder holds an encoding.Encoder for encoding events.
+	Encoder Encoder
+
+	// Sync can be used to indicate whether production should be synchronous.
+	Sync bool
+
+	// Mutators holds the list of RecordMutator applied to all the records sent
+	// by the producer.
```
marclop, line 67:
```suggestion
	// by the producer. If any errors are returned, the producer will not
	// produce and return the error in ProcessBatch.
```

kafka/producer.go
```diff
@@ -17,4 +17,170 @@
...
+
+// Validate checks that cfg is valid, and returns an error otherwise.
+func (cfg ProducerConfig) Validate() error {
+	if cfg.Broker == "" {
+		return errors.New("broker cannot be empty")
+	}
+	if cfg.Logger == nil {
+		return errors.New("logger cannot be nil")
+	}
+	return nil
```
marclop, line 79: I missed this in the previous review: Could you prefix the errors with the name of the package and use Go `1.20` semantics to be able to return a _multi-error_? Similar to what's being done here: https://github.com/elastic/apm-queue/blob/7e52ea6dc17dd6c60dbd6fd62651cc11c94997f2/kafka/consumer.go#L64-L85.

We're also missing a nil check on the cfg.Encoder. Could you add that as well, so caller's don't cause a panic?

marclop (commented):
Almost there! I missed the Validate() method in my first couple of reviews. After these comments are addressed, we're good to merge.

kafka/producer.go
kruskall: Thanks for mentioning this! Updated!

kafka/producer.go
```diff
@@ -17,4 +17,176 @@
...
+func (cfg ProducerConfig) Validate() error {
+	var err []error
+	if cfg.Broker == "" {
+		err = append(err, errors.New("broker cannot be empty"))
+	}
+	if cfg.Logger == nil {
+		err = append(err, errors.New("logger cannot be nil"))
+	}
+	if cfg.Encoder == nil {
+		err = append(err, errors.New("encoder cannot be nil"))
```
marclop, line 83: I think we should prefix the messages with the package name to keep it consistent with the current producer.
```suggestion
		err = append(err, errors.New("kafka: broker cannot be empty"))
	}
	if cfg.Logger == nil {
		err = append(err, errors.New("kafka: logger cannot be nil"))
	}
	if cfg.Encoder == nil {
		err = append(err, errors.New("kafka: encoder cannot be nil"))
```

marclop (approved):
LGTM!

kafka/producer.go
kruskall: Thanks! Somehow I missed that part :facepalm:

PR #9, opened 2023-02-28
context: Enrich / extract metadata (string map)

marclop:
Removes the project context enrichment / extraction and instead uses a more versatile and abstract map[string]string to store arbitrary k/v in the context.

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -158,32 +158,30 @@ func (c *Consumer) fetch(ctx context.Context) error {
 		)
 	})
 	fetches.EachRecord(func(msg *kgo.Record) {
-		ctx := context.Background()
-		for _, h := range msg.Headers {
-			if h.Key == "project_id" {
-				ctx = queuecontext.WithProject(ctx, string(h.Value))
-				break
-			}
-		}
 		var event model.APMEvent
 		if err := c.cfg.Decoder.Decode(msg.Value, &event); err != nil {
 			// TODO(marclop) DLQ?
-			c.cfg.Logger.Error("unable to unmarshal json into model.APMEvent",
+			c.cfg.Logger.Error("unable to decode message.Value into model.APMEvent",
 				zap.Error(err),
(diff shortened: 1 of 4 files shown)


pubsublite/consumer.go
```diff
@@ -125,11 +125,10 @@ func NewConsumer(ctx context.Context, cfg ConsumerConfig) (*Consumer, error) {
 		NackHandler: func(msg *pubsub.Message) error {
 			// TODO(marclop) DLQ?
 			partition, offset := partitionOffset(msg.ID)
-			projectID := msg.Attributes["project_id"]
 			cfg.Logger.Error("handling nacked message",
 				zap.Int("partition", partition),
 				zap.Int64("offset", offset),
-				zap.String("project_id", projectID),
+				zap.Any("project_id", msg.Attributes),
```
axw, line 131:
```suggestion
				zap.Any("attributes", msg.Attributes),
```

kafka/consumer.go
```diff
@@ -158,32 +158,30 @@ func (c *Consumer) fetch(ctx context.Context) error {
...
 		if err := c.cfg.Decoder.Decode(msg.Value, &event); err != nil {
 			// TODO(marclop) DLQ?
-			c.cfg.Logger.Error("unable to unmarshal json into model.APMEvent",
+			c.cfg.Logger.Error("unable to decode message.Value into model.APMEvent",
 				zap.Error(err),
 				zap.String("topic", msg.Topic),
 				zap.ByteString("message.value", msg.Value),
 				zap.Int64("offset", msg.Offset),
-				zap.Int32("partition", int32(msg.Partition)),
+				zap.Int32("partition", msg.Partition),
```
axw, line 169: should we log the headers here, like attributes are logged for pubsublite?

axw (approved):

pubsublite/consumer.go
marclop: 🤦🏼‍♂️

PR #11, opened 2023-03-02
pubsublite(consumer): default AtMostOnce delivery

marclop:
Introduce a `Delivery` configuration option for the pubsublite consumer, allowing the caller to configure the message acknowledgement strategy.

The two supported strategies are: `AtMostOnce` and `AtLeastOnce`. The former is the current default (if left unspecfied), while the latter is not exactly at least once (depending on the implementation of the batch processor).

The configuration options have been documented to avoid confusion.

### DIFF
--- a/pubsublite/consumer.go
+++ b/pubsublite/consumer.go
@@ -195,9 +195,9 @@ func (c *Consumer) Run(ctx context.Context) error {
 		case apmqueue.AtLeastOnceDeliveryType:
 			defer func() {
 				if err != nil {
-					defer msg.Nack()
+					msg.Nack()
 				} else {
-					defer msg.Ack()
+					msg.Ack()
 				}
 			}()
 		}
--- a/pubsublite/consumer.go
+++ b/pubsublite/consumer.go
@@ -29,21 +29,9 @@ import (
 	"google.golang.org/api/option"
 
 	"github.com/elastic/apm-data/model"
+	apmqueue "github.com/elastic/apm-queue"
 	"github.com/elastic/apm-queue/queuecontext"
 )
-
-const (
-	// AtMostOnceDeliveryType acknowleges the message as soon as it's received
-	// and decoded, without waiting for the message to be processed.
-	AtMostOnceDeliveryType DeliveryType = iota
-	// AtLeastOnceDeliveryType acknowledges the message after it has been
-	// processed. It may or may not create duplicates, depending on how batches
-	// are processed by the underlying model.BatchProcessor.
-	AtLeastOnceDeliveryType
-)
-
-// DeliveryType for the consumer. For more details See the supported DeliveryTypes.
-type DeliveryType uint8
 
 // Decoder decodes a []byte into a model.APMEvent
 type Decoder interface {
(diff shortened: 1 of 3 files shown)


pubsublite/consumer.go
```diff
@@ -32,6 +32,19 @@ import (
...
 
+const (
+	// AtMostOnceDeliveryType acknowleges the message as soon as it's received
+	// and decoded, without waiting for the message to be processed.
+	AtMostOnceDeliveryType DeliveryType = iota
+	// AtLeastOnceDeliveryType acknowledges the message after it has been
+	// processed. It may or may not create duplicates, depending on how batches
+	// are processed by the underlying model.BatchProcessor.
+	AtLeastOnceDeliveryType
+)
```
lahsivjar, line 43: Should we have the `DeliveryType` for all queues?

axw: Yes, IIUC franz-go's default is to auto-commit the prior poll's records, on each poll. We could keep that for AtLeastOnce, and otherwise disable autocommit and call CommitUncommittedOffsets prior to calling Fetches.EachRecord.

marclop: I can move the DeliveryType to `apmqueue` and remove the constructors that we have there since I don't think they're useful and we won't use them much. That way, we can reuse these in all the implementations.

axw: ++ SGTM. If we want to reintroduce the constructor later, it could be in a separate package that imports pubsublite & kafka

marclop: moved the DeliveryType to the root package.

pubsublite/consumer.go
```diff
@@ -178,8 +188,20 @@ func (c *Consumer) Run(ctx context.Context) error {
...
+		var err error
+		switch c.cfg.Delivery {
+		case apmqueue.AtMostOnceDeliveryType:
+			msg.Ack()
+		case apmqueue.AtLeastOnceDeliveryType:
+			defer func() {
+				if err != nil {
+					defer msg.Nack()
+				} else {
+					defer msg.Ack()
```
axw, line 200:
```suggestion
					msg.Nack()
				} else {
					msg.Ack()
```

queue.go
```diff
@@ -21,35 +21,22 @@ package apmqueue
...
-	"github.com/elastic/apm-queue/pubsublite"
 )
 
 const (
-	_ QueueType = iota
-	// QueueTypeKafka defines the kafka queue type
-	QueueTypeKafka QueueType = iota
-	// QueueTypePubSubLite defines the PubSub Lite queue type.
-	QueueTypePubSubLite QueueType = iota
+	// AtMostOnceDeliveryType acknowleges the message as soon as it's received
```
axw, line 29:
```suggestion
	// AtMostOnceDeliveryType acknowledges the message as soon as it's received
```

axw (approved):

pubsublite/consumer.go
axw: Opened https://github.com/elastic/apm-queue/issues/12 for kafka

pubsublite/consumer.go
marclop: if we want at least once delivery, then we should acknowledge **after** the message has been processed, not before, right? Same for Nack.

marclop: oh never mind, double defer 🤦🏼‍♂️

PR #13, opened 2023-03-02
pubsublite(consumer): Receive from N subscriptions

marclop:
Adds support for receiving messages from multiple subscriptions at once with the same configuration.


axw (approved):
LGTM! Perhaps we should add a comment to ConsumerConfig.Processor that Processor may be called concurrently by multiple goroutines?

PR #14, opened 2023-03-02
pubsublite(producer): Produce to multiple topics

marclop:
Supports producing events to multiple topics based on an EventRouter that is set by the client.

### DIFF
--- a/pubsublite/producer.go
+++ b/pubsublite/producer.go
@@ -38,15 +38,21 @@ type Encoder interface {
 	Encode(model.APMEvent) ([]byte, error)
 }
 
+// RouterFunc returns the topic to use for a given APMEvent.
+type RouterFunc func(model.APMEvent) Topic
+
 // ProducerConfig for the PubSub Lite producer.
 type ProducerConfig struct {
-	// Topic is the PubSub Lite topic.
-	Topic Topic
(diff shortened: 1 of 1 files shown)


pubsublite/producer.go
```diff
@@ -38,15 +38,21 @@ type Encoder interface {
...
+	// Topics are the PubSub Lite topics where the messages will be produced.
+	// The routing is determined by the EventRouter.
+	Topics []Topic
 	// Encoder holds an encoding.Encoder for encoding events.
 	Encoder Encoder
 	// Logger for the producer.
-	Logger     *zap.Logger
-	ClientOpts []option.ClientOption
+	Logger *zap.Logger
+	// EventRouter returns the topic where an event should be produced.
```
axw, line 53:
```suggestion
	// EventRouter returns the topic where an event should be produced.
	//
	// If EventRouter returns a value that does not exist in Topics, then
	// ProcessBatch will return an error.
```

Alternatively we could lazily create producers?  This is fine for now though.

axw (approved):

pubsublite/producer.go
marclop: yeah I suppose we could do it lazily, maybe we can create an issue to improve this? it may require some unlocking and re-locking I'm not to keen on for now.

PR #17, opened 2023-03-07
producer: Introduce TopicRouter

marclop:
Introduce a new `apmqueue.TopicRouter`, used in both implementations of the producer to route a given event to a topic.

The `pubsublite.Producer` no longer groups the topics by project and region, and those are settings need to be set only once.


queue.go
```diff
@@ -56,3 +56,9 @@ type Producer interface {
 	// Close closes the producer.
 	Close() error
 }
+
+// Topic represents a destination topic where to produce a message/record.
+type Topic string
+
+// TopicRouter is used to determine the destinatin topic for an model.APMEvent.
```
axw, line 63:
```suggestion
// TopicRouter is used to determine the destination topic for an model.APMEvent.
```

axw (approved):
LGTM :)

PR #21, opened 2023-03-09
Use slice of broker addresses in kafka producer

carsonip:
Breaking change for kafka producer to accept brokers instead of just one broker.


marclop (approved):
Thanks!

PR #22, opened 2023-03-09
consumer(kafka): Default to AtMostOnceDelivery

### LINKED ISSUE #12: consumer(Kafka): Add support for `DeliveryType`
https://github.com/elastic/apm-queue/pull/11#discussion_r1122611441

marclop:
## Description
Changes the default delivery mechanism for the Kafka consumer from at least once to at most once delivery. This aligns the consumer with the `pubsublite` counterpart.

Additionally, bounds the maximum number of records to for a single poll, and makes it configurable through a configuration option.

## Related issues

Closes #12


kafka/consumer.go
```diff
@@ -102,6 +109,15 @@ func NewConsumer(cfg ConsumerConfig) (*Consumer, error) {
...
 		kgo.WithLogger(kzap.New(cfg.Logger)),
 	}
+	switch cfg.Delivery {
+	case apmqueue.AtLeastOnceDeliveryType:
+		// franz-go defaults to at least once processing, using auto committing.
+	case apmqueue.AtMostOnceDeliveryType:
+		// GreedyAutoCommit opts in to committing everything that has been polled
+		// when autocommitting (the dirty offsets), rather than committing what
+		// has previously been polled.
+		opts = append(opts, kgo.GreedyAutoCommit())
```
axw, line 119: Auto-commit seems to be done asynchronously, in a background goroutine: https://github.com/twmb/franz-go/blob/3322f311fd833129f91b4ea76581150274bdb457/pkg/kgo/consumer_group.go#L2009

So I don't think setting GreedyAutoCommit is enough to guarantee at most once delivery. I think we would need to disable autocommit, and manually commit before consuming the fetches.

marclop: ah, you're right. I'll switch to manual commit then.

marclop: I've made the required changes and added blocking rebalances during record a poll and until `client.AllowRebalance()` is called (to avoid committing offsets for partitions lost).

kafka/consumer.go
```diff
@@ -142,16 +168,34 @@ func (c *Consumer) Run(ctx context.Context) error {
...
 func (c *Consumer) fetch(ctx context.Context) error {
 	// NOTE(marclop) this is pretty naive consuming, to maximize throughput,
 	// it's best to use one goroutine per partition, but that requires more
 	// state management and blocking when rebalances happen.
 	c.mu.RLock()
 	defer c.mu.RUnlock()
-	fetches := c.client.PollFetches(ctx)
+
+	fetches := c.client.PollRecords(ctx, c.cfg.MaxPollRecords)
+	defer c.client.AllowRebalance()
```
axw, line 181: Should we be allowing rebalance immediately after committing offsets?

axw (approved):
LGTM, just wondering if we should call AllowRebalance directly after CommitUncommittedOffsets for AtMostOnceDeliveryType, and directly after polling for AtLeastOnceDeliveryType. Maybe we should only use BlockRebalanceOnPoll with AtMostOnceDeliveryType?

kafka/consumer.go
marclop: I suppose we could since we've marked the records as committed. I was trying to avoid calling `c.client.AllowRebalance()` in different places, I can make so if you think it's best:

```go
fetches := c.client.PollRecords(ctx, c.cfg.MaxPollRecords)
if fetches.IsClientClosed() || errors.Is(fetches.Err0(), context.Canceled) {
	return context.Canceled // Client closed or context cancelled.
}
switch c.cfg.Delivery {
case apmqueue.AtLeastOnceDeliveryType:
	// Commit the fetched record offsets after we've processed them.
	defer func() {
		c.client.CommitUncommittedOffsets(ctx)
		c.client.AllowRebalance()
	}()
case apmqueue.AtMostOnceDeliveryType:
	// Commit the fetched record offsets as soon as we've polled them.
	err := c.client.CommitUncommittedOffsets(ctx)
	c.client.AllowRebalance()
	if err != nil {
		// If the commit fails, then return immediately and any uncommitted
		// records will be re-delivered in time. Otherwise, records may be
		// processed twice.
		return nil
	}
}
```

I think using `BlockRebalanceOnPoll` is still advantageous to try and reduce duplicates for `AtLeastOnceDeliveryType`, since it guards against that scenario.

axw: IIANM `AllowRebalance` is idempotent, so how about this?

```go
fetches := c.client.PollRecords(ctx, c.cfg.MaxPollRecords)
defer c.client.AllowRebalance()
...
switch c.cfg.Delivery {
case apmqueue.AtLeastOnceDeliveryType:
	// Commit the fetched record offsets after we've processed them.
	defer c.client.CommitUncommittedOffsets(ctx)
case apmqueue.AtMostOnceDeliveryType:
	// Commit the fetched record offsets as soon as we've polled them.
	if err := c.client.CommitUncommittedOffsets(ctx); err != nil {
		// If the commit fails, then return immediately and any uncommitted
		// records will be re-delivered in time. Otherwise, records may be
		// processed twice.
		return nil
	}
	// Allow rebalancing now that we have committed offsets, preventing
	// another consumer from reprocessing the records.
	c.client.AllowRebalance()
}
...
```

marclop: LGTM. Thanks!

PR #26, opened 2023-03-12
refactor: update logging and error messages for consistency

kruskall:
format and make sure zap field enrichment and logged messages are consistent


kafka/consumer.go
```diff
@@ -228,6 +231,7 @@ func (c *Consumer) fetch(ctx context.Context) error {
 			c.cfg.Logger.Error("unable to process event",
 				zap.Error(err),
 				zap.String("topic", msg.Topic),
+				zap.ByteString("message.value", msg.Value),
```
kruskall, line 234: **question**: logging of headers and message values seem to differ wildly. When should they be logged ? Does it make sense to always log them ?

kafka/consumer.go
```diff
@@ -228,6 +231,7 @@ func (c *Consumer) fetch(ctx context.Context) error {
...
 				zap.Int64("offset", msg.Offset),
 				zap.Int32("partition", msg.Partition),
 				zap.Any("headers", meta),
```
kruskall, line 237: **question**: should we use a more general name for the log field ? `metadata` maybe ?

kafka/consumer.go
```diff
@@ -180,8 +180,11 @@ func (c *Consumer) fetch(ctx context.Context) error {
...
-	if fetches.IsClientClosed() || errors.Is(fetches.Err0(), context.Canceled) {
-		return context.Canceled // Client closed or context cancelled.
+	if fetches.IsClientClosed() {
+		return fmt.Errorf("client is closed: %w", context.Canceled)
+	}
+	if errors.Is(fetches.Err0(), context.Canceled) {
+		return fmt.Errorf("context canceled: %w", fetches.Err0())
 	}
 	switch c.cfg.Delivery {
 	case apmqueue.AtLeastOnceDeliveryType:
```
kruskall, line 190: **question**: we are not checking for error when committing uncommited offsets. Should we log an error or/and return the error to the caller ?

kafka/consumer.go
```diff
@@ -109,7 +109,7 @@ type Consumer struct {
 // NewConsumer creates a new instance of a Consumer.
 func NewConsumer(cfg ConsumerConfig) (*Consumer, error) {
 	if err := cfg.Validate(); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("invalid consumer config: %w", err)
```
marclop, line 112:
```suggestion
		return nil, fmt.Errorf("kafka: invalid consumer config: %w", err)
```

kafka/consumer.go
marclop: No need, since this function is running on a loop. 

If the offsets are committed back to Kafka, the events will be re-processed once the processing timeout is hit, adhering to the desired at least once delivery.

kafka/consumer.go
marclop: I added the message value in the log above since that error occurs on decoding a message. Processing the batch may happen for any reason and it seemed to me that it wasn't necessary to log the raw `message.value`. I still think it's best to not log the message.value wherever possible.

kafka/producer.go
```diff
@@ -181,6 +181,10 @@ func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch) error {
 				p.cfg.Logger.Error("failed producing message",
 					zap.Error(err),
 					zap.String("topic", msg.Topic),
+					zap.ByteString("message.value", msg.Value),
```
marclop, line 184: I'd like to avoid logging the `message.value` unless necessary for debugging.
```suggestion
```

pubsublite/consumer.go
```diff
@@ -133,7 +133,7 @@ type Consumer struct {
 // NewConsumer creates a new consumer instance for a single subscription.
 func NewConsumer(ctx context.Context, cfg ConsumerConfig) (*Consumer, error) {
 	if err := cfg.Validate(); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("invalid consumer config: %w", err)
```
marclop, line 136:
```suggestion
		return nil, fmt.Errorf("pubsublite: invalid consumer config: %w", err)
```

pubsublite/consumer.go
```diff
@@ -253,8 +254,10 @@ func (c consumer) processMessage(ctx context.Context, msg *pubsub.Message) {
 		partition, offset := partitionOffset(msg.ID)
 		c.logger.Error("unable to process event",
 			zap.Error(err),
+			zap.ByteString("message.value", msg.Data),
```
marclop, line 257:
```suggestion
```

pubsublite/producer.go
```diff
@@ -96,7 +96,7 @@ type Producer struct {
 // NewProducer creates a new PubSub Lite producer for a single project.
 func NewProducer(ctx context.Context, cfg ProducerConfig) (*Producer, error) {
 	if err := cfg.Validate(); err != nil {
-		return nil, err
+		return nil, fmt.Errorf("invalid producer config: %w", err)
```
marclop, line 99:
```suggestion
		return nil, fmt.Errorf("pubsublite: invalid producer config: %w", err)
```

pubsublite/producer.go
```diff
@@ -161,17 +161,23 @@ func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch) error {
...
-	for _, res := range responses {
-		if serverID, err := res.Get(ctx); err != nil {
-			p.cfg.Logger.Error("failed producing message",
-				zap.Error(err),
-				zap.String("server_id", serverID),
-			)
-		}
+		result := producer.Publish(ctx, &msg)
+		wg.Add(1)
+		go func() {
```
marclop, line 166: This PR is about updating logging and error messages for consistency, this seems a much greater change than that. Could you revert these changes, please?

marclop (changes requested):

kafka/consumer.go
marclop: ideally both Kafka and pubsublite should have the same field. Either headers or metadata seem fine to me.

kafka/consumer.go
kruskall: Oh I see, thanks for explanation! :bow:

kafka/consumer.go
kruskall: Oh I see, I've updated the code to remove the message.value during processing errors.

Should the headers also be removed ? Are they valuable ?

pubsublite/producer.go
kruskall: Ah, this was to avoid storing tmp variables to retrieve topics and headers later on. I've changed it to a tmp variable

kafka/consumer.go
kruskall: Let's keep headers then!

kafka/consumer.go
marclop: I think it's fine to log the headers, and may be useful to be able to identify failing patterns for given metadata.

marclop (approved):
LGTM. Thank you!

PR #34, opened 2023-03-21
consumer(kafka): Concurrently consume partitions

### LINKED ISSUE #23: consumer(Kafka): Process records concurrently
## Description

The current consumer implementation polls records from the Kafka broker and serially process them in a single goroutine. This is almost certainly going to result in reduced throughput (although it is dependent on the implementation of the `model.BatchProcessor` which passed to the consumer).

marclop:
Updates the Kafka consumer to consume from partitions via dedicated go- routines. Each assigned partition for the consumer in the consumer group will be consumed concurrently.

A `consumer` and `partitionConsumer` types are introduced, the former manages the assigned and revoked / lost partitions, and ensure that the latter (`partitionConsumer`) is created on assignment of new partitions, and stopped on revoked / lost partitions.

Closes #23

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -125,7 +125,7 @@ func NewConsumer(cfg ConsumerConfig) (*Consumer, error) {
 		kgo.SeedBrokers(cfg.Brokers...),
 		kgo.ConsumerGroup(cfg.GroupID),
 		kgo.ConsumeTopics(cfg.Topics...),
-		kgo.WithLogger(kzap.New(cfg.Logger.Named("franz"))),
+		kgo.WithLogger(kzap.New(cfg.Logger.Named("kafka"))),
 		// If a rebalance happens while the client is polling, the consumed
 		// records may belong to a partition which has been reassigned to a
 		// different consumer int he group. To avoid this scenario, Polls will
@@ -172,7 +172,7 @@ func (c *Consumer) Close() error {
 	// Since we take the full lock when closing, there's no need to explicitly
 	// allow rebalances since polls aren't concurrent with Close().
 	c.client.Close()
-	c.consumer.close() // Close each partition consumer.
+	c.consumer.wg.Wait() // Wait for all the goroutines to exit.
 	return nil
 }
 
@@ -226,7 +226,10 @@ func (c *Consumer) fetch(ctx context.Context) error {
 			return
 		}
 		tp := topicPartition{topic: ftp.Topic, partition: ftp.Partition}
-		c.consumer.consumers[tp].records <- ftp.Records
+		select {
+		case c.consumer.consumers[tp].records <- ftp.Records:
+		case <-ctx.Done():
+		}
 	})
 	return nil
 }
@@ -244,6 +247,7 @@ func (c *Consumer) Healthy() error {
 // to use when partitions are reassigned.
 type consumer struct {
 	mu        sync.Mutex
+	wg        sync.WaitGroup
 	consumers map[topicPartition]partitionConsumer
 	processor model.BatchProcessor
 	logger    *zap.Logger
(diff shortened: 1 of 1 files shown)


kafka/consumer.go
```diff
@@ -156,6 +172,7 @@ func (c *Consumer) Close() error {
 	// Since we take the full lock when closing, there's no need to explicitly
 	// allow rebalances since polls aren't concurrent with Close().
 	c.client.Close()
+	c.consumer.close() // Close each partition consumer.
```
axw, line 175: does `Client.Close` not invoke the partition revoked callbacks?

kafka/consumer.go
```diff
@@ -249,3 +239,165 @@ func (c *Consumer) Healthy() error {
...
+	decoder   Decoder
+	delivery  apmqueue.DeliveryType
+}
+
+// consume processed the records from a topic and partition. Calling consume
+// more than once will cause a panic.
+func (pc partitionConsumer) consume(topic string, partition int32) {
+	defer func() {
+		close(pc.done)
+		close(pc.records)
```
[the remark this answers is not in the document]
marclop: Thanks! That seems much cleaner to implement.

kafka/consumer.go
```diff
@@ -249,3 +239,165 @@ func (c *Consumer) Healthy() error {
...
+		for _, partition := range partitions {
+			tp := topicPartition{topic: topic, partition: partition}
+			pc := c.consumers[tp]
+			delete(c.consumers, tp)
+			close(pc.stop)
+			wg.Add(1)
+			go func() { <-pc.done; wg.Done() }()
+		}
+	}
+	wg.Wait()
```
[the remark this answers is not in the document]
marclop: I was reading the code again and I don't think it's necessary since at this point the polling is stopped (rebalance is happening), and once the entry is removed from the map, it isn't accessible by TopicPartition anymore.

kafka/consumer.go
marclop: > ... but we don't do need that, because we're explicitly committing even for at-least-once?

Correct, for at-least-once, each goroutine is responsible to commit the processed offsets. For at-most-once, we commit ahead of time, so again there's no need to commit on the revoked callback.

> does Client.Close not invoke the partition revoked callbacks?

You're right, I missed this bit:

> Internally, this function leaves the group, and leaving a group causes a rebalance so that you can get one final notification of revoked partitions

I can remove the `c.consumer.close()` call since it's redundant.

kafka/consumer.go
```diff
@@ -207,35 +220,15 @@ func (c *Consumer) fetch(ctx context.Context) error {
...
-				zap.String("topic", msg.Topic),
-				zap.Int64("offset", msg.Offset),
-				zap.Int32("partition", msg.Partition),
-				zap.Any("headers", meta),
-			)
-			return
+		tp := topicPartition{topic: ftp.Topic, partition: ftp.Partition}
+		select {
+		case c.consumer.consumers[tp].records <- ftp.Records:
+		case <-ctx.Done():
```
axw, line 231: Could you please add a comment to the `<-ctx.Done()` case about interaction with at least/most once delivery?

kafka/consumer.go
```diff
@@ -249,3 +242,143 @@ func (c *Consumer) Healthy() error {
...
+			pc := partitionConsumer{
+				wg:        &c.wg,
+				records:   make(chan []*kgo.Record),
+				processor: c.processor,
+				logger:    c.logger,
+				decoder:   c.decoder,
+				client:    client,
+				delivery:  c.delivery,
+			}
+			go pc.consume(topic, partition)
```
axw, line 280:
```suggestion
			go func(topic string, partition int32) {
				defer c.wg.Done()
				pc.consume(topic, partition)
			}(topic, partition)
```

Then the partition consumer doesn't need to know about the wait group, and it's kept close to where you call Add.

axw (approved):
LGTM!

kafka/consumer.go
marclop: done!

PR #35, opened 2023-03-21
producer(kafka): Add basic produce test

marclop:
Adds a basic test for the Kafka producer using the newly added `kfake` package from `franz-go`. Only basic producing and consuming is supported for now.


axw (approved):
Very nice!

Issue #37, opened 2023-03-22
Kafka: Allow TLS and SASL to be configured


marclop:
## Description

The current Kafka producer and consumer only supports PLAIN un-encrypted communication with the Kafka cluster. In production, it's highly likely that the Kafka cluster is configured to use TLS for authentication and / or SASL for authorization. This issue tracks the required work to add it to the library.

PR #39, opened 2023-03-23
kafka(sasl): Abstract sasl.Mechanism via packages

marclop:
Abstracts the underlying `sasl.Mechanism` interface (originating from `franz-go`) in favor of dedicated `sasl/<type>` packages. A new package `sasl/plain` is introduced, and `sasl/aws` is introduced as a module to avoid importing the AWS SDK directly in AWS queue for cases when the importer may not want or need to use the AWS MSK IAM sasl type.

Part of #37


kruskall (approved):
LGTM

PR #56, opened 2023-03-29
action: export GH workflow runs to APM data

v1v:
This will help to visualise the GitHub workflow runs as APM traces

It requires a couple of secrets to be added, though I'm not admin, so I can send you over them 

Cheers


axw (approved):
> It requires a couple of secrets to be added, though I'm not admin, so I can send you over them

I have made observablt-robots admins

PR #58, opened 2023-03-30
kafka(producer): Fix async producing

### LINKED ISSUE #53: kafka(producer): Allow producer to produce asynchronously
## Description

The current producer code will return immediately if when the producing isn't synchronous, which may cause the records to fail to be produced if the context is cancelled before the records are produced to Kafka. To avoid this situation. We should detach the context deadline from the current context if the production isn't synchronous.

## Code

https://github.com/elastic/apm-queue/blob/ea9b407c0406556a985fecb629a9e9074f0f4322/kafka/producer.go#L208-L211

marclop:
Fixes the async producing by detaching the `context.Context.Deadline` from the passed context in ProcessBatch, but maintaining context.Values that may be defined, which allows trace context to be propagated down to the kgo client.

Closes #53

### DIFF
--- a/kafka/producer.go
+++ b/kafka/producer.go
@@ -184,7 +184,7 @@ func (p *Producer) Close() error {
 
 // ProcessBatch publishes the batch to the kafka topic inferred from the
 // configured TopicRouter. If the Producer is synchronous, it waits until all
-// messages have been produced to PubSub Lite, otherwise, returns as soon as
+// messages have been produced to Kafka, otherwise, returns as soon as
 // the messages have been stored in the producer's buffer.
 func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch) error {
 	// Take a read lock to prevent Close from closing the client
--- a/kafka/producer.go
+++ b/kafka/producer.go
@@ -171,6 +171,10 @@ func NewProducer(cfg ProducerConfig) (*Producer, error) {
 }
 
 // Close stops the producer
+//
+// This call is blocking and will cause all the underlying clients to stop
+// producing. If producing is asynchronous, it'll block until all messages
+// have been produced. After Close() is called, Producer cannot be reused.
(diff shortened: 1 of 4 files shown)


kafka/producer.go
```diff
@@ -171,14 +171,21 @@ func NewProducer(cfg ProducerConfig) (*Producer, error) {
...
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	p.client.Close()
 	return nil
 }
 
-// ProcessBatch publishes the events in batch to the specified Kafka topic.
+// ProcessBatch publishes the batch to the kafka topic inferred from the
+// configured TopicRouter. If the Producer is synchronous, it waits until all
+// messages have been produced to PubSub Lite, otherwise, returns as soon as
```
axw, line 187:
```suggestion
// messages have been produced to Kafka, otherwise, returns as soon as
```

kafka/producer.go
```diff
@@ -212,6 +219,10 @@ func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch) error {
 			return fmt.Errorf("failed to encode event: %w", err)
 		}
 		record.Value = encoded
+		if !p.cfg.Sync {
+			// Detach the context from its deadline or cancellation.
+			ctx = queuecontext.DetachedContext(ctx)
```
axw, line 224: Probably a stupid question, but why do we need this? I would think that if the client closes the connection we should abort the publish.

axw (commented):
Looks good, just wondering why we need this

kafka/producer.go
axw: Never mind, I think I understand now.

marclop: It's got to do with the `client.Publish(ctx)` actually happening after the context may have been cancelled. We had a similar problem in the APM Server Intake calls with async set.

Issue #42, opened 2023-03-23
Testing infrastructure


axw:
We should create some testing infrastructure, primarily for use during development:

- A command for publishing events to a queue (perhaps receiving ND-JSON as input?) (#72)
- ~A command for consuming events from a queue (#85)~
- System / integration tests that run against the two queue implementations.
  - System test framework (#92)
    - Infrastructure provisioning, wait for infrastructure to be available.
    - Tear down after tests are complete or with an additional command.
  - Basic producing (#93)
    - Synchronous and Asynchronous producing
    - Topic routing
  - Group consuming (#94)
~- Run system tests as part of PR checks (approved contributors) (#95).~

simitt:
I moved https://github.com/elastic/apm-queue/issues/95 out of scope for `ms-03`. Let's discuss that with the robots team when this can be tackled. 

Please ensure to get https://github.com/elastic/apm-queue/issues/94 merged within the `ms-03`, so that we can close this meta issue.

cachedout:
@simitt Could you please provide more detail on what you would like our team to do in this regard?

simitt:
@cachedout that comment was refering to #95, where we already discussed that the robots team is taking on the automation (ss https://github.com/elastic/apm-queue/issues/95#issuecomment-1527036259)

simitt:
Moved this task to `ms-04`, the PR looks like it is at a good stage to get merged in this iteration.

marclop:
@simitt is this task also meant to track the CI automation side of things? The initial tests have been all added now that #116 is closed.

PR #78, opened 2023-04-10
Add `loadgen` module with helm chart

marclop:
Adds a new loadgen module with its helm chart to allow easy deployment to a k8s cluster.

Part of #42


queue.go
```diff
@@ -62,3 +63,36 @@ type Topic string
 
 // TopicRouter is used to determine the destination topic for an model.APMEvent.
 type TopicRouter func(event model.APMEvent) Topic
+
+// NewEventTypeTopicRouter returns a topic router which routes events based on
+// an optional prefix, and the event.Processor.Event field for an event.
+func NewEventTypeTopicRouter(prefix string) TopicRouter {
```
lahsivjar, line 69: Does it make sense to keep this in the loadgen package or do you forsee this being used elsewhere too?

lahsivjar (approved):
LGTM!

queue.go
marclop: I think we're likely to reuse it for other producers.

Issue #71, opened 2023-04-05
kafka(consumer): Add consumer group unit tests


marclop:
## Description

The `kfake` module recently added support for group consuming, joining, etc. We should add tests to assert consumption work as expected.

PR #97, opened 2023-04-18
test: add kafka consumer unit tests

kruskall:
Add consumer unit tests: 

- health check
- config
- consume events
- multiple consumers 

Coverage for the consumer code is now at 83.8%:
- `go test -v -coverprofile=profile.txt ./...`
- `go tool cover -html profile.txt`

Closes https://github.com/elastic/apm-queue/issues/71


kafka/consumer_test.go
```diff
@@ -18,12 +18,238 @@
...
+		Logger:  zap.NewNop(),
+		Processor: model.ProcessBatchFunc(func(ctx context.Context, b *model.Batch) error {
+			count.Add(1)
+			assert.Len(t, *b, 1)
+			//assert.Equal(t, event, (*b)[0])
+			return nil
+		}),
+	}
+
+	b, err := json.JSON{}.Encode(event)
```
marclop, line 191: Let's use the same codec for producing and consuming. That'll make it easier to swap in the future too without changing multiple lines.

```suggestion
	codec := json.JSON{}

	cfg := ConsumerConfig{
		Brokers: addrs,
		Topics:  topics,
		GroupID: "groupid",
		Decoder: codec,
		Logger:  zap.NewNop(),
		Processor: model.ProcessBatchFunc(func(_ context.Context, b *model.Batch) error {
			assert.Len(t, *b, 1)
			assert.Equal(t, event, (*b)[0])
			return nil
		}),
	}

	b, err := codec.Encode(event)
```

kafka/consumer_test.go
```diff
@@ -18,12 +18,238 @@
...
+		Decoder: json.JSON{},
+		Logger:  zap.NewNop(),
+		Processor: model.ProcessBatchFunc(func(ctx context.Context, b *model.Batch) error {
+			assert.Len(t, *b, 1)
+			assert.Equal(t, event, (*b)[0])
+			return nil
+		}),
+	}
+
+	b, err := json.JSON{}.Encode(event)
```
marclop, line 152: Same comment here ont he codec.

marclop (commented):
Tests look good! Would you mind adding two more tests?

- TestMultipleConsumerGroups (Assert that each consumer group gets a copy of all records)

kafka/consumer_test.go
```diff
@@ -18,12 +18,305 @@
...
+	cfg := ConsumerConfig{
+		Brokers: addrs,
+		Topics:  topics,
+		GroupID: "groupid",
+		Decoder: codec,
+		Logger:  zap.NewNop(),
+		Processor: model.ProcessBatchFunc(func(ctx context.Context, b *model.Batch) error {
+			count.Add(1)
+			assert.Len(t, *b, 1)
+			//assert.Equal(t, event, (*b)[0])
```
marclop, line 189: Should remove this line
```suggestion
```

kafka/consumer_test.go
```diff
@@ -18,12 +18,305 @@
...
+			})
+
+		}
+	}()
+
+	<-waitCh
+
+	for k, i := range m {
+		assert.Eventually(t, func() bool {
+			return i.Load() == int64(amount)
```
marclop, line 291: It's not necessary to unit test 100 different consumer groups and tie the number of consumer groups to the number of produced records. I think it would be better to have two variables, one for the records produced and another for the number of consumer groups. Also there's no need to launch a goroutine for the producer and sync with a channel:

```suggestion
	groups := 10
	for i := 0; i < groups; i++ {
		gid := "groupid" + strconv.Itoa(i)
		cfg.GroupID = gid
		m[gid] = &atomic.Int64{}
		cfg.Processor = model.ProcessBatchFunc(func(ctx context.Context, b *model.Batch) error {
			count := m[gid]
			count.Add(1)
			assert.Len(t, *b, 1)
			assert.Equal(t, event, (*b)[0])
			return nil
		})
		consumer := newConsumer(t, cfg)
		go func() {
			assert.ErrorIs(t, consumer.Run(ctx), context.Canceled)
		}()
	}

	produceRecords := 100
	for i := 0; i < produceRecords; i++ {
		client.Produce(context.Background(), &kgo.Record{
			Topic: topics[0],
			Value: b,
		}, func(r *kgo.Record, err error) {
			assert.NoError(t, err)
			assert.NotNil(t, r)
		})
	}

	for k, i := range m {
		assert.Eventually(t, func() bool {
			return i.Load() == int64(produceRecords)
```

marclop (approved):
Looks great, thanks!

One thing I would like to have in the unit tests is consumption from multiple topics. Right now, we're only testing consumption from a single topic. It would be great to have multiple topic tests in there too. If you wouldn't mind doing it as a follow-up or writing up an issue so we don't forget.

kruskall:
Done: https://github.com/elastic/apm-queue/issues/107

PR #98, opened 2023-04-18
Add OpenTelemetry hook to kafka consumer

dmathieu:
This adds the OpenTelemetry hook to the kafka consumer


marclop (approved):
Looks good. I wonder if we should start the instrumentation in `ProcessBatch`, since there's some work being done there.

dmathieu:
I have added a trace to ProcessBatch.

kruskall (approved):

PR #100, opened 2023-04-18
pubsublite: Add OpenTelemetry instrumentation

dmathieu:
This adds instrumentation for the pubsublite consumer.


pubsublite/internal/telemetry/publisher.go
```diff
@@ -0,0 +1,66 @@
...
+
+	if msg.Attributes == nil {
+		msg.Attributes = make(map[string]string)
+	}
+	otel.GetTextMapPropagator().Inject(ctx, propagation.MapCarrier(msg.Attributes))
+
+	res := h(ctx, msg)
+
+	go func() {
+		mid, err := res.Get(ctx)
```
marclop, line 55: I don't like a couple of things about this approach:

- `res.Get()` is called twice, here and in `pubsublite/producer.go`.
- A goroutine is created for each published message. This may create quite a lot of overhead in the Go runtime for mid-high throughput producers.

I don't have a great alternative for this right now, but maybe worth dropping a NOTE in the code?

marclop (commented):
Looks pretty good! I'm concerned about some potential goroutine explosion for tracing and what that's going to mean for applications that have instrumentation turned on.

pubsublite/internal/telemetry/publisher.go
```diff
@@ -0,0 +1,66 @@
...
+		trace.WithSpanKind(trace.SpanKindProducer),
+		trace.WithAttributes(
+			semconv.MessagingSystemKey.String("pubsub"),
+			semconv.MessagingDestinationKindTopic,
+		),
+	)
+
+	if msg == nil {
+		msg = &pubsub.Message{}
+	}
```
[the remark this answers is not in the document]
dmathieu: Yeah, it shouldn't happen. It was more to be thorough in tests.

pubsublite/internal/telemetry/consumer_test.go
```diff
@@ -0,0 +1,171 @@
...
+				spans[i].Parent = spans[i].Parent.
+					WithTraceID(tt.traceID).
+					WithSpanID(tt.parentSpanID).
+					WithTraceFlags(0)
+				spans[i].Resource = nil
+				spans[i].StartTime = time.Time{}
+				spans[i].EndTime = time.Time{}
+			}
+
+			assert.Equal(t, spans, tt.expectedSpans)
```
kruskall, line 156:
```suggestion
			assert.Equal(t, tt.expectedSpans, spans)
```

pubsublite/internal/telemetry/consumer_test.go
```diff
@@ -0,0 +1,171 @@
...
+	"go.opentelemetry.io/otel/sdk/instrumentation"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/sdk/trace/tracetest"
+	semconv "go.opentelemetry.io/otel/semconv/v1.18.0"
+	"go.opentelemetry.io/otel/trace"
+)
+
+func TestConsumer(t *testing.T) {
+	exp := tracetest.NewInMemoryExporter()
+	tp := sdktrace.NewTracerProvider(
```
kruskall, line 40: **suggestion**: we should shutdown the tracer to release any resource

pubsublite/internal/telemetry/consumer_test.go
```diff
@@ -0,0 +1,171 @@
...
+			exp.Reset()
+		})
+	}
+}
+
+func mustTraceIDFromHex(s string) (t trace.TraceID) {
+	var err error
+	t, err = trace.TraceIDFromHex(s)
+	if err != nil {
+		panic(err)
```
kruskall, line 168: **suggestion**: should we use testify `require.NoError` for consistency ?

pubsublite/internal/telemetry/consumer_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+			assert.Equal(t, spans, tt.expectedSpans)
+
+			// Reset for next tests
+			exp.Reset()
+		})
+	}
+}
+
+func mustTraceIDFromHex(s string) (t trace.TraceID) {
```
kruskall, line 164: issue: we need a `t.Helper` here to print the correct line number in case of failure

pubsublite/internal/telemetry/publisher_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"go.opentelemetry.io/otel/sdk/instrumentation"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	"go.opentelemetry.io/otel/sdk/trace/tracetest"
+	semconv "go.opentelemetry.io/otel/semconv/v1.18.0"
+	"go.opentelemetry.io/otel/trace"
+)
+
+func TestPublisher(t *testing.T) {
+	exp := tracetest.NewInMemoryExporter()
+	tp := sdktrace.NewTracerProvider(
```
kruskall, line 41: **issue**: same as above, we should shutdown the tracer

pubsublite/internal/telemetry/publisher_test.go
```diff
@@ -0,0 +1,160 @@
...
+				spans[i].Resource = nil
+				spans[i].StartTime = time.Time{}
+				spans[i].EndTime = time.Time{}
+
+				for k := range spans[i].Events {
+					spans[i].Events[k].Time = time.Time{}
+				}
+			}
+
+			assert.Equal(t, spans, tt.expectedSpans)
```
kruskall, line 154:
```suggestion
			assert.Equal(t, tt.expectedSpans, spans)
```

kruskall (changes requested):

pubsublite/internal/telemetry/consumer.go
```diff
@@ -0,0 +1,60 @@
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/propagation"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+	"go.opentelemetry.io/otel/trace"
+)
+
+type consumerHandler = func(context.Context, *pubsub.Message)
+
+// Consumer adds telemetry data to messages received
+func Consumer(tracer trace.Tracer, h consumerHandler) consumerHandler {
```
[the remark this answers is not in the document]
dmathieu: I'm adding those attributes.
I've chosen to add an attributes parameter, even though it feels a bit weird to expose otel to `consumer.go`. I could also add the subscription to the consumer struct, so we get the additional information within the tracer.

pubsublite/internal/telemetry/publisher.go
```diff
@@ -0,0 +1,66 @@
...
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/propagation"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+	"go.opentelemetry.io/otel/trace"
+)
+
+type producerHandler = func(context.Context, *pubsub.Message) *pubsub.PublishResult
+
+// Publisher adds telemetry data to messages published
+func Publisher(ctx context.Context, tracer trace.Tracer, msg *pubsub.Message, h producerHandler) *pubsub.PublishResult {
```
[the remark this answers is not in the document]
dmathieu: As attributes, or elsewhere? I've added name, project and region as attributes.

pubsublite/internal/telemetry/publisher_test.go
```diff
@@ -0,0 +1,160 @@
...
+					Status: sdktrace.Status{
+						Code:        codes.Error,
+						Description: "context canceled",
+					},
+				},
+			},
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			res := &pubsub.PublishResult{}
```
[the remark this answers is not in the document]
dmathieu: The thing is that it's a channel we have to close. And `reflect` will not us call [Close](https://pkg.go.dev/reflect#Value.Close) on a channel retrieved from a private value 😢

pubsublite/internal/telemetry/publisher.go
marclop: I also came across: https://github.com/googleapis/google-cloud-go/issues/2953. Perhaps we can add this link to the comment so we've a bit more context?

Would you mind opening an issue in this repo about the potential problems of this approach? One way to possibly mitigate this would be to return the span in the Result, and handle the end of the span lifecycle as a callback function. We're already handling the responses back in the `pubsublite/producer.go`, so calling an extra function when the response is obtained may not be a bad thing to do. The referenced issue has some details on how messages are batched and sent too.

pubsublite/internal/telemetry/publisher.go
```diff
@@ -0,0 +1,68 @@
...
+	ctx, span := tracer.Start(ctx, "pubsublite.Publish",
+		trace.WithSpanKind(trace.SpanKindProducer),
+		trace.WithAttributes(
+			semconv.MessagingSystemKey.String("pubsub"),
```
marclop, line 38:
```suggestion
			semconv.MessagingSystemKey.String("pubsublite"),
```

pubsublite/internal/telemetry/publisher.go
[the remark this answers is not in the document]
marclop: Yes as an attribute, [`semconv.MessagingDestinationNameKey`](https://pkg.go.dev/go.opentelemetry.io/otel@v1.14.0/semconv/v1.18.0#MessagingDestinationNameKey) seems appropriate. Where have you added the region and project in the publisher? I see you added them in the consumer, but not the publisher.

pubsublite/consumer.go
```diff
@@ -171,6 +175,11 @@ func NewConsumer(ctx context.Context, cfg ConsumerConfig) (*Consumer, error) {
 				zap.String("region", subscription.Region),
 				zap.String("project", subscription.Project),
 			),
+			telemetryAttributes: []attribute.KeyValue{
+				semconv.MessagingSourceNameKey.String(subscription.Name),
+				attribute.String("region", subscription.Region),
```
marclop, line 180: Shouldn't this be using [`semconv.CloudRegion`](https://pkg.go.dev/go.opentelemetry.io/otel@v1.14.0/semconv/v1.18.0#CloudRegion)? 

```suggestion
				semconv.CloudRegion(subscription.Region),
```

pubsublite/consumer.go
```diff
@@ -171,6 +175,11 @@ func NewConsumer(ctx context.Context, cfg ConsumerConfig) (*Consumer, error) {
...
+				attribute.String("project", subscription.Project),
```
marclop, line 181:
```suggestion
				semconv.CloudAccountID(subscription.Project),
```

pubsublite/internal/telemetry/publisher_test.go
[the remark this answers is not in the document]
marclop: Ah, I missed that, well then, not much we can do

pubsublite/internal/telemetry/publisher.go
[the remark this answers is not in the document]
dmathieu: Ah, yes, sorry. I had missed the publisher. I just added it.

pubsublite/internal/telemetry/publisher_test.go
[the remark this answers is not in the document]
dmathieu: I've opened an issue: https://github.com/googleapis/google-cloud-go/issues/7801

kruskall (approved):
LGTM!

pubsublite/internal/telemetry/publisher_test.go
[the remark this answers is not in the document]
marclop: Thanks!

marclop (commented):
The kafka and this implementation seem to differ slightly in what can be configured:

In Kafka we've got a `TraceProvider trace.TracerProvider` and `DisableTelemetry bool`. Wouldn't it be better if both packages accepted the same flags to enable and configure the instrumentation?

dmathieu:
I've added a `TracerProvider` config.
For `DisableTelemetry`, I'm not sure. I added it in case the hook had unintended side effects. In the case of manual telemetry though, we own the code and can disable telemetry globally by not configuring an SDK.

Similar discussion: https://github.com/elastic/apm-queue/pull/104#discussion_r1173292307

marclop (approved):

Issue #15, opened 2023-03-02
pubsublite(producer): Lazily provision unknown topics returned by EventRouter


marclop:
```suggestion
	// EventRouter returns the topic where an event should be produced.
	//
	// If EventRouter returns a value that does not exist in Topics, then
	// ProcessBatch will return an error.
```

Alternatively we could lazily create producers?  This is fine for now though.

_Originally posted by @axw in https://github.com/elastic/apm-queue/pull/14#discussion_r1122742432_

PR #101, opened 2023-04-19
feat: lazily provision unknown topics returned by EventRouter

kruskall:
lazily provision unknown topics.

Keep the old `Topics` option and abstract topic provisioning into a separate method for consistency

Close https://github.com/elastic/apm-queue/issues/15


pubsublite/producer.go
```diff
@@ -202,7 +190,11 @@ func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch) error {
 		topic := p.cfg.TopicRouter(event)
 		producer, ok := p.producer[topic]
 		if !ok {
-			return fmt.Errorf("pubsublite: unable to find producer for %s", topic)
+			publisher, err := newPublisher(ctx, p.cfg, topic)
+			if err != nil {
+				return fmt.Errorf("pubsublite: failed creating publisher: %w", err)
+			}
+			p.producer[topic] = publisher
```
marclop, line 197: This method acquires a read lock, yet it mutates the producer map, which will lead to races when this method is called concurrently.

Looking at the implementation and how we use the map, I think it may be best to use a `sync.Map`, so we don't have to introduce a new Mutex.

marclop (changes requested):

pubsublite/producer.go
kruskall: Good catch! Updated :+1:

marclop (commented):
Thanks for the changes! The core approach looks good.

Since we now allow topics to be created on the fly (or lazily), I think it would make more sense to also remove the requirement to pass in a list of topics, and update the `Config.Topics` field's good to reflect the changes we've performed.

pubsublite/producer.go
```diff
@@ -200,16 +189,25 @@ func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch) error {
 			}
 		}
 		topic := p.cfg.TopicRouter(event)
-		producer, ok := p.producer[topic]
+		// creating a publisher is expensive and we can't do it lazily so we call Load+LoadOrStore
```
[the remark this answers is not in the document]
kruskall: Sorry, I wasn't clear. In this specific context (`sync.Map` access) we can't create a value lazily:

What I have in mind for lazily creating a value is something like
```
map.LoadOrStore(key, func() any { 
  // lazily create the value if missing
})
```

(java has something similar, it's optimized and does 1 lookup, creating the value lazily if missing)

So we need to use `Load`+`LoadOrStore` (accessing the map twice) otherwise a single `LoadOrStore(topic, newpublisher())` wuould create a publisher (unused) for each batch.

pubsublite/producer.go
```diff
@@ -110,7 +105,7 @@ type resTopic struct {
 type Producer struct {
 	mu        sync.RWMutex
 	cfg       ProducerConfig
-	producer  map[apmqueue.Topic]*pscompat.PublisherClient
+	producer  sync.Map // map[apmqueue.Topic]*pscompat.PublisherClient
```
dmathieu, line 108: I know this isn't something new to this PR. But it's weird to have a slice names singularly. How about renaming it to `producers`?

dmathieu (commented):
This seems like something where unit tests (especially ones to check for race conditions) could be useful?

marclop (commented):
Code looks good. I would love to see unit tests for this, but AFAICS we can't add them easily since we're using the upstream types and there's no mock available to use provided by Google. I've opened an issue in https://github.com/googleapis/google-cloud-go/issues/7842, to see if the existing pubsublite/test package can be made available so we can use it.

Alternatively, we could abstract the client building and introduce an interface that complies with the current `pscompat.PublisherClient`, but that is a bit more involved.

@dmathieu Did you have any other ideas on how to accomplish better unit testing?

dmathieu:
Hmm no, based on the lack of testing capabilities from the Google APIs, I'm afraid there isn't much we can do here :(

PR #104, opened 2023-04-20
Ensure tracing properly happened in kafka ProcessBatch

dmathieu:
Following https://github.com/elastic/apm-queue/pull/102, this adds assertions checking that the manual span is indeed created and stopped.


kafka/producer_test.go
```diff
@@ -117,6 +129,22 @@ func TestNewProducerBasic(t *testing.T) {
 			//lint:ignore SA1012 passing a nil context is a valid use for this call.
 			fetches := client.PollRecords(nil, 1)
 			assert.Len(t, fetches.Records(), 0)
+
+			// Assert tracing happened properly
+			assert.Equal(t, len(exp.GetSpans()), spanCount+1)
```
marclop, line 134: Shouldn't there also be a span from the `franz-go` library?

dmathieu: Hmm probably. I'll look into that.
Also, the async test is somehow failing in CI, but not locally.

kafka/producer.go
```diff
@@ -197,8 +203,10 @@ func (p *Producer) Close() error {
 // the messages have been stored in the producer's buffer.
 func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch) error {
 	ctx, span := p.tracer.Start(ctx, "producer.ProcessBatch", trace.WithAttributes(
+		attribute.Bool("sync", p.cfg.Sync),
 		attribute.Int("batch.size", len(*batch)),
 	))
+	defer span.End()
```
dmathieu, line 209: So, I know I've brought this back and reverted #102.
But there were 2 issues with that PR:

We can't end the span within `Produce`, because it runs once for each event, which means the span is closed after the first event instead of the last.

If this method is async, it effectively returns quick, even though things are still running but in the background. (child spans can still be created, they will just show up as running after this method finished). Waiting for every child span to finish instead of the method to return gives a wrong visualization of what's happening.

kafka/producer_test.go
dmathieu: Indeed, kotel wasn't enabled at all due to a configuration issue. I've fixed it.

kafka/producer.go
```diff
@@ -236,10 +249,8 @@ func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch) error {
...
 		p.client.Produce(ctx, record, func(msg *kgo.Record, err error) {
-			defer func() {
-				wg.Done()
-				span.End()
-			}()
+			defer wg.Done()
+
 			if err != nil {
 				span.RecordError(err)
 				span.SetStatus(codes.Error, err.Error())
```
kruskall, line 256: **question**: if the span ended and production is performed async, can we still change the status and record the error here ? I don't think that's possible and we might need a child span

kafka/producer_test.go
marclop: Thanks for looking into it.

kafka/producer.go
marclop: Should we also add the k/v Paris from `queuecontext.MetadataFromContext`? There may be useful contextual information in those.

kafka/producer.go
```diff
@@ -236,10 +249,8 @@ func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch) error {
 			ctx = queuecontext.DetachedContext(ctx)
 		}
 		p.client.Produce(ctx, record, func(msg *kgo.Record, err error) {
```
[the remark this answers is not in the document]
marclop: If we do, then we should probably not use the `kotel` tracer, since it won't add much value? we'd need to duplicate some of attribute setting here.

kafka/producer.go
```diff
@@ -197,8 +209,10 @@ func (p *Producer) Close() error {
 // the messages have been stored in the producer's buffer.
 func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch) error {
 	ctx, span := p.tracer.Start(ctx, "producer.ProcessBatch", trace.WithAttributes(
```
marclop, line 211: It seems we're always starting a trace, even when `cfg.DisableTelemetry` is  set.

marclop (commented):
Maybe not for this PR, but the `ConsumerConfig` doesn't seem to have a `TracerProvider trace.TracerProvider`.

kafka/producer.go
dmathieu: This could be fixed with a `Produce` span, which tracks our own method (not the kgo internal ones, so it's different).
Then, the span isn't finished when this happens async.

kafka/producer.go
dmathieu: Yeah, I setup `DisableTelemetry` thinking the kotel hook may have unintended side effects.
Disabling manual tracing can be done by not setting up an otel SDK though.

We could rename this to `DisableKgoTelemetry`, or entirely remove it?

kafka/producer.go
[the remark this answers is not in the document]
dmathieu: I don't know. They both track slightly different things. See also https://github.com/elastic/apm-queue/pull/104#discussion_r1173372090

dmathieu:
The consumer doesn't do any manual tracing. So it doesn't need a tracer provider. We can add one when we add manual tracing to it.

kafka/producer.go
dmathieu: We definitely can. I'm always a bit wary of adding attributes without knowing what they are.
That metadata never includes any PII or sensitive data?

kafka/producer.go
marclop: We'd probably want to call `span.End()` for each of the `Produce` spans on the `Produce callback` that `kgo` calls after the record is produced.

dmathieu: Isn't that what we already do?
Now that you mention it though, I wonder if we should even track errors in this callback. kgo will already record the error in its own hook.

So couldn't we entirely ignore that error?

kafka/producer.go
```diff
@@ -251,10 +264,13 @@ func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch) error {
 			ctx = queuecontext.DetachedContext(ctx)
 		}
 		p.client.Produce(ctx, record, func(msg *kgo.Record, err error) {
-			defer func() {
-				wg.Done()
-				span.End()
-			}()
+			_, span := p.tracer.Start(ctx, "producer.Produce", trace.WithAttributes(
```
marclop, line 267: This doesn't seem right to me, the span is called after the event has been produced to Kafka, so what are we tracking here?

What could be an acceptable approach would be to start the span in line `236`, so we have a span tracking the time it takes to create and produce a `kgo.Record` individually, and also can set the correct state, and track the production errors if any. That's why I was implying that the kgo hooks won't add much value then.

dmathieu: And as you mentioned, that would be re-tracing what kotel does.
Kotel also already handles errors in its hook, hence the re-tracing too.

I've removed this span, and added a comment about it. Is that ok with you?

marclop (approved):
LGTM!

PR #105, opened 2023-04-20
fix(go.mod): update loadgen go.mod and go.sum

simitt:
Update `go.mod` and `go.sum` by running `go mod tidy`. This is required to successfully run the loadgenerator, also when used in the tilt setup in `apm-managed-service`.


endorama (approved):
I checked out this branch to fix my tilt setup and it worked as expected, `loadgen` service is correctly started.

PR #108, opened 2023-04-24
Sleep in producer test to reduce the risk of flaky test

dmathieu:
franz-go runs the hooks to finish events in an unchecked goroutine:

https://github.com/twmb/franz-go/blob/master/pkg/kgo/producer.go#L436-L440

This method calls `finishPromises`, which calls `OnProduceRecordUnbuffered`, which is the hook that kotel uses to end the span.

Because of that, the check on the number of created spans is flaky: https://github.com/elastic/apm-queue/actions/runs/4763014463/jobs/8466352708

We don't have any way to know whether kgo has finished running the goroutine. So by sleeping for a millisecond, we reduce the risk of the flaky issue happening.
If kgo took more than 1ms to process the hooks, that seems like something which would be a performance issue to investigate anyway.

### DIFF
--- a/kafka/producer_test.go
+++ b/kafka/producer_test.go
@@ -130,13 +130,11 @@ func TestNewProducerBasic(t *testing.T) {
 			fetches := client.PollRecords(nil, 1)
 			assert.Len(t, fetches.Records(), 0)
 
-			// kgo runs an unchecked goroutine to mark all the spans it creates as
-			// finished (and close the spans). So the only way to ensure it ran is to
-			// wait. Without this, we get flaky tests in CI.
-			time.Sleep(time.Millisecond)
-
 			// Assert tracing happened properly
-			assert.Equal(t, spanCount+3, len(exp.GetSpans()))
+			assert.Eventually(t, func() bool {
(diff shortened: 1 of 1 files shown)


kafka/producer_test.go
```diff
@@ -130,8 +130,13 @@ func TestNewProducerBasic(t *testing.T) {
 			fetches := client.PollRecords(nil, 1)
 			assert.Len(t, fetches.Records(), 0)
 
+			// kgo runs an unchecked goroutine to mark all the spans it creates as
+			// finished (and close the spans). So the only way to ensure it ran is to
+			// wait. Without this, we get flaky tests in CI.
+			time.Sleep(time.Millisecond)
```
simitt, line 136: How about moving this into a loop with eg N=3 (there might be higher delays on CI) and starting with the `assert.Equal` check before the sleep?

dmathieu: Good idea. Your comment actually reminded me that testify has an `assert.Eventually`, which is much better here.
https://pkg.go.dev/github.com/stretchr/testify/assert#Eventually

PR #119, opened 2023-05-02
pubsublite: configure consumer with topics

axw:
Replace ConsumerConfig.Subscriptions with
Project, Region, and Topics. These are used
by the implementation to form a subscription.

This brings the kafka and pubsublite APIs in line.

### DIFF
--- a/pubsublite/consumer.go
+++ b/pubsublite/consumer.go
@@ -47,8 +47,12 @@ type Decoder interface {
 
 // ConsumerConfig defines the configuration for the PubSub Lite consumer.
 type ConsumerConfig struct {
-	// PubSub Lite subscriptions.
-	Subscriptions []Subscription
+	// Region is the GCP region for the producer.
+	Region string
+	// Project is the GCP project for the producer.
+	Project string
+	// Topics holds Pub/Sub Lite topics from which messages will be consumed.
+	Topics []apmqueue.Topic
 	// Decoder holds an encoding.Decoder for decoding events.
 	Decoder Decoder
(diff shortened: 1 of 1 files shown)


pubsublite/consumer.go
```diff
@@ -163,8 +168,13 @@ func NewConsumer(ctx context.Context, cfg ConsumerConfig) (*Consumer, error) {
...
+	consumers := make([]consumer, 0, len(cfg.Topics))
+	for _, topic := range cfg.Topics {
+		subscription := Subscription{
+			Name:    string(topic),
+			Project: cfg.Project,
+			Region:  cfg.Region,
+		}
 		client, err := pscompat.NewSubscriberClientWithSettings(
 			ctx, subscription.String(), settings, cfg.ClientOpts...,
 		)
```
dmathieu, line 180: Should we change the `logger` and `telemetryAttributes` down below to use the data from cfg rather than the subscription?

axw: Seems equivalent to me. Do you see a benefit in doing that?

dmathieu: It is indeed equivalent. The only benefit I would see is in the ease of reading/understanding. But that's very subjective.

axw: Pushed https://github.com/elastic/apm-queue/pull/119/commits/4fb881b6af890a664aeb25d548cf765762589c1d, is that what you had in mind? I don't personally see an improvement, but if you find this easier to read and understand then that works for me :)

dmathieu (approved):

PR #121, opened 2023-05-03
pubsublite: fix lazy loading, remove Topics

axw:
- Topics is not needed if we lazily create publisher clients on first use.
- Don't use the request context when lazily creating a publisher
- Close the non-stored publisher on race to store, return the stored one


axw:
I have run the systemtests locally, all passing.

marclop (approved):

systemtest/produce_consume_basic_test.go
```diff
@@ -169,17 +170,18 @@ func TestProduceConsumeMultipleTopics(t *testing.T) {
...
 				}),
 				consumer: newPubSubLiteConsumer(ctx, t, pubsublite.ConsumerConfig{
-					Logger:        logger,
-					Decoder:       json.JSON{},
-					Subscriptions: pubSubLiteSubscriptions(topics...),
+					Logger:  logger,
+					Decoder: json.JSON{},
+					Topics:  topics,
+					Project: googleProject,
+					Region:  googleRegion,
```
marclop, line 184: These two will always need to be set to `googleProject` and `googleRegion`. Should we do that in `newPubSubLiteConsumer`?

axw: SGTM, will update

Issue #123, opened 2023-05-03
kafka: Extend unit tests


marclop:
## Description

By leveraging the `kfake` package, we can provide pretty good test coverage for the unit tests. We should extend the current code coverage and add a few more tests in these areas:

### Producer

- Lossless graceful shutdown (No records should be lost when the producer is closed), both async and sync.

### Consumer

- At Most Once Delivery (default).
- At Least Once Delivery.
- Lossless graceful shutdown (No records should be lost when the consumer is closed).

kruskall:
I have finished the Lossless graceful shutdown test and I have a few questions:

> (No records should be lost when the consumer is closed)

did you mean `when the producer is closed` ? I assumed that was a typo and I am closing the producer in the test. 

___

Could you clarify what we want to test with lossless graceful shutdown ?

This is the current test for the consumer graceful shutdown: 

https://github.com/elastic/apm-queue/blob/5e50011b202e23f36f222dff572af84ab2b5b9b6/kafka/consumer_test.go#L372

We produced some recorded and we're waiting for the consumer to process them. Once that's done we cancel the context:

https://github.com/elastic/apm-queue/blob/5e50011b202e23f36f222dff572af84ab2b5b9b6/kafka/consumer_test.go#L409-L415

The processor is unblocking the select by reading from the channel and incrementing the counter:

https://github.com/elastic/apm-queue/blob/5e50011b202e23f36f222dff572af84ab2b5b9b6/kafka/consumer_test.go#L389-L396

At that point the final assertion will make sure the counter match the produced recors:

https://github.com/elastic/apm-queue/blob/5e50011b202e23f36f222dff572af84ab2b5b9b6/kafka/consumer_test.go#L416

But this will always succeed because the processor was called and incremented the counter. If the processor is not called the test will timeout early.
My test is similar, except it is using the producer to create records. Apologies if I'm missing something obvious, I feel like the logic of tests is "verify that by closing the consumer after the consumer processed events, we are not losing events". I can't see that logic/test fails, wouldn't it always succeed ? How does shutdown relates to that ? Wouldn't shutdown happens in the middle of processing ? Or even before it ?

kruskall:
Thank you for the clarification! :bow:

kruskall:
I think this can be closed. All the tests mentioned in the description should be present.

@marclop Am I Missing something ?

Issue #118, opened 2023-05-02
kafka(consumer): Document/clarify the consumer behavior, or implement re-processing if processor fails


marclop:
## Description

Based on the guidance given in https://github.com/twmb/franz-go/issues/440#issuecomment-1528899269. We should decide what our `AtLeastOnceDelivery` guarantees are. Currently, it may be expected that a consumer tries to run the passed Processor more than once if processing fails, but that is not the case. 

`AtLeastOnceDelivery` may only occur if a consumer has fetched some records from Kafka, but the process is terminated, crashes or loses network connectivity. After the consumer misses enough check-ins, the Kafka broker will consider that consumer instance in the consumer group as dead, and will resign its assigned partitions to other active consumers.

However, if the configured `Processor` fails to process a single record, the commit is skipped and the next Fetch may pull new records, commit the new offset and will lead to up to `MaxPollRecords` being lost for a partition.

https://github.com/elastic/apm-queue/blob/a482794ff7551d00b828ec866e8e0d060e97a253/kafka/consumer.go#L76-L82

https://github.com/elastic/apm-queue/blob/a482794ff7551d00b828ec866e8e0d060e97a253/kafka/consumer.go#L64-L69

#### Implemented in #125

The simplest path forward would be to continue processing the records that can be processed instead of stopping the processing of partitions record at that offset and potentially losing the rest of that fetch and log those that cannot be. This will lead to less data loss for `AtMostOnceDelivery`. Update the documentation accordingly.

### 2. Implement some sort of per-partition processing retries

Instead of dropping the events that can't be processed, a per-partition sized buffer could be kept which could be retried on a timer or when there aren't any active records fetched for the partition.

This option increases the memory footprint of the consumer and is slightly more complicated to implement well.

axw:
Since we're not planning to use `AtLeastOnceDelivery` in the near future, I think (1) is preferable for now. Longer term I think we should restart the consumer on error, rewind to the last committed offset + 1, and start processing again; this might mean some events are reprocessed.

lahsivjar:
How about using a DLQ for this? We can have a single DLQ for all messages, a DLQ consumer will consume from the DLQ with a delay and put it back to the original topic for another retry.

marclop:
Solution 1 has been implemented as part of #125.

[...]

PR #125, opened 2023-05-03
kafka(consumer): Improve `AtLeastOnce` handling

marclop:
Improves how `AtLeastOnceDelivery` is handled by the kafka consumer. Previously, the consumer could discard up to `cfg.MaxPollRecords`, since it incorrectly assumed that unacknowledged records would be re-delivered by Kafka if not committed within a certain amount of time.

Now, it discards any records which can't be processed, but continues processing the rest of the records in the fetch. It can still result in some data loss, and is at the mercy of the failures returned by the processor, but it's much better than previously.

Additionally, adds a few tests which cover both `AtLeastOnceDelivery` and `AtMostOnceDelivery`, and a `TestGracefulShutdown` test.

Partially addresses #118
Part of #123


kruskall (approved):

marclop:
Ran the tests a lot of times locally and 10 times in the CI to verify that there's not any flaky tests left: https://github.com/elastic/apm-queue/actions/runs/4890005702.

PR #116, opened 2023-04-27
systemtest: add group consuming tests

kruskall:
As mentioned, the "at least once" tests are failing.

Add systemtests to test different delivery types. 

Test is sending 2 batches of events (100 + 100) with the first event erroring out during processing.

AtMostOnce works correctly and we get 199 events.
AtLeastOnce has inconsistent behaviour between kafka and pubsublite.

Related to https://github.com/elastic/apm-queue/issues/94

PR #126, opened 2023-05-08
kafka(consumer): Fix At Least Once delivery commit

marclop:
## Description

Update the partition consumer to skip committing offsets of records that have failed processing when AtLeastOnceDelivery is used. May not cause re-processing of records if newer records are produced to the partition, unless the consumer crashes or the partition is reassigned to an another consumer in the consumer group. This is still preferrable to committing the offset of a failed record, given that AtLeastOnceDelivery is set.

## Related issues

Fixes Kafka failures in #116


axw (approved):

kruskall (approved):
LGTM!

PR #127, opened 2023-05-08
pubsublite(consumer): Fix AtLeastOnce re-delivery

marclop:
## Description
Fixes the previously Acked / Nacked failed to process events. Implements a simple dead-lettering of 2 processing attempts, after which the event will be nacked and logged.

## Related issues

Fixes the PubSub Lite failures from #116

### DIFF
--- a/pubsublite/consumer.go
+++ b/pubsublite/consumer.go
@@ -125,7 +125,7 @@ func (cfg ConsumerConfig) Validate() error {
 type Consumer struct {
 	mu             sync.Mutex
 	cfg            ConsumerConfig
-	consumers      []consumer
+	consumers      []*consumer
 	stopSubscriber context.CancelFunc
 	tracer         trace.Tracer
 }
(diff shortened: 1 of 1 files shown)


pubsublite/consumer.go
```diff
@@ -264,11 +275,29 @@ func (c consumer) processMessage(ctx context.Context, msg *pubsub.Message) {
...
+				if attempt > 2 {
+					msg.Nack()
+					c.failed.Delete(msg.ID)
+					return
+				}
+				c.failed.Store(msg.ID, attempt)
+				return
 			}
+			partition, offset := partitionOffset(msg.ID)
+			c.logger.Info("processed previously failed event",
```
axw, line 294: IIANM the event may alternatively have been processed on the first attempt?

axw (approved):

marclop:
@kruskall That is because the consumer will stop consuming after the context is cancelled (after 60s). The timeout needs to be increased to at least 90s, possibly better to 120s.

The first message is consumed in 35-40s (This is because the PubSub consumer is a bit slow to initialize, docs point to up to 60s for initialization, after that it's low-latency high throughput), and the second one is consumed in the same timeline, which means that it would consume the unacknowledged message in at least 70-80s.

kruskall:
Understood! Thanks for the details! :bow:

PR #141, opened 2023-05-16
pubsublite(consumer): Add minimal logging

marclop:
Adds a bit more logging to the PubSub Lite consumer. This should make it easier to spot any processing errors and re-processing attempts.

Added while working on fixing #116

### DIFF
--- a/pubsublite/consumer.go
+++ b/pubsublite/consumer.go
@@ -201,7 +201,10 @@ func NewConsumer(ctx context.Context, cfg ConsumerConfig) (*Consumer, error) {
 func (c *Consumer) Close() error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	c.stopSubscriber()
+	if c.stopSubscriber != nil {
+		c.cfg.Logger.Info("closing pubsublite consumer...")
+		c.stopSubscriber()
+	}
 	return nil
 }
 
@@ -228,6 +231,7 @@ func (c *Consumer) Run(ctx context.Context) error {
 				))
 				// Keep attempting to receive until a fatal error is received.
 				if errors.Is(err, pscompat.ErrBackendUnavailable) {
+					c.cfg.Logger.Info("transient error found, re-starting consumer...", zap.Error(err))
 					continue
 				}
 				return err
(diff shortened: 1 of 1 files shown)


pubsublite/consumer.go
```diff
@@ -201,7 +201,10 @@ func NewConsumer(ctx context.Context, cfg ConsumerConfig) (*Consumer, error) {
 func (c *Consumer) Close() error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	c.stopSubscriber()
+	if c.stopSubscriber != nil {
```
marclop, line 204: If the consumer hadn't been started, close would panic.

pubsublite/consumer.go
```diff
@@ -283,10 +287,12 @@ func (c *consumer) processMessage(ctx context.Context, msg *pubsub.Message) {
 					attempt += a.(int)
 				}
 				if attempt > 2 {
+					c.logger.Info("re-processing limit exceeded, discarding event", zap.Int("attempt", attempt))
```
marclop, line 290: Should this be a warn?

axw: I think so

pubsublite/consumer.go
```diff
@@ -228,6 +231,7 @@ func (c *Consumer) Run(ctx context.Context) error {
 				))
 				// Keep attempting to receive until a fatal error is received.
 				if errors.Is(err, pscompat.ErrBackendUnavailable) {
+					c.cfg.Logger.Info("transient error found, re-starting consumer...", zap.Error(err))
```
dmathieu, line 234: We should have a span here. So we could also fetch it from context and add an event to it (so it's easier to search and match to a specific event)

marclop: it seems strange to me to have a span here. `Receive()` may block for hours or days if no transient or fatal errors occur. Is this what you had in mind?

dmathieu: Ah no, right. Somehow, I thought it returned the received object. Yeah, in that case, it makes sense. There shouldn't be any spans (we could create one to pass the log message, but that'd just be a different way to do logging).

Issue #124, opened 2023-05-03
kafka: Extend system tests


marclop:
## Description

This issue should covers the work required to implement the test cases in #123 as system tests

Issue #122, opened 2023-05-03
pubsublite: Add comprehensive system tests


marclop:
## Description

We don't have many pubsublite tests for either the consumer or producer. This is due to limitations in the current Google SDK. To ensure we have adequate code coverage, we must add system tests that test the following areas:

### Consumer

  - At Most Once Delivery (default).
  - At Least Once Delivery.
  - Lossless graceful shutdown (No records should be lost when the consumer is closed).

### Producer
  - Lazy creation of clients and asserting production to of records to topics.
  - Lossless graceful shutdown (No records should be lost when the consumer is closed), both async and sync.

## Related issues

- https://github.com/googleapis/google-cloud-go/issues/7842
- https://github.com/googleapis/google-cloud-go/issues/7801

PR #143, opened 2023-05-16
test: add systemtests for lossless graceful shutdown

kruskall:
Add lossless graceful shutdown systemtests.

Mostly inspired by the current lossless graceful shutdown unit tests.

Related to https://github.com/elastic/apm-queue/issues/124
Related to https://github.com/elastic/apm-queue/issues/122


marclop (commented):
There's a lot of duplicate code, would you mind refactoring the test body so that it both kafka and pubsublite do the same without copy pasting the test code?

kruskall:
> There's a lot of duplicate code, would you mind refactoring the test body so that it both kafka and pubsublite do the same without copy pasting the test code?

Agree with you but since this seems to be carried over from other tests, I'll do that in a followup issue

PR #152, opened 2023-05-18
Introduce common config structs

axw:
Introduce `kafka.CommonConfig` and `pubsublite.CommonConfig`, taking the common config fields from their consumer and producer config structs of their respective packages. Just a little refactoring to reduce duplication and ensure consistency between consumers and producers.


pubsublite/producer.go
```diff
@@ -117,24 +99,17 @@ func NewProducer(cfg ProducerConfig) (*Producer, error) {
...
 		cfg:    cfg,
 		closed: make(chan struct{}),
 		// NOTE(marclop) should the channel size be dynamic? 1000 is an arbitrary
 		// number, but it must be greater than 0, so async produces don't block.
 		responses: make(chan []resTopic, 1000),
-		tracer:    tracerProvider.Tracer("pubsublite"),
+		tracer:    cfg.tracerProvider().Tracer("pubsublite"),
 
 		project: cfg.Project,
 		region:  cfg.Region,
```
lahsivjar, line 111: We don't seem to be using these 2 fields anymore.

lahsivjar (approved):
LGTM, minor comment.

pubsublite/producer.go
axw: Indeed, thanks. Probably should set up staticcheck at some point.

PR #162, opened 2023-05-24
{kafka,pubsublite}: add Manager types

axw:
Add types for managing the creation of Kafka topics, and the creation of Pub/Sub Lite reservations, topics, and subscriptions. System tests will be updated to use these in a followup PR.

pubsublite.ConsumerConfig has been updated to replace Topics with Subscriptions. A topic may have multiple subscriptions, roughly equivalent to consumer groups in Kafka.


kafka/manager.go
```diff
@@ -0,0 +1,204 @@
...
+						"failed to create topic %q: %w",
+						response.Topic, err,
+					),
+				)
+			}
+			continue
+		}
+		logger.Info("created kafka topic", createTopicParamsFields...)
+	}
+	if err := errors.Join(createErrors...); err != nil {
```
dmathieu, line 156: This is condition seems to essentially be:

```
if (err != nil) { return err } else { return nil }
```

Which is the equivalent of `return err` (if it's nil, we'll return nil anyway).
So the condition seems redundant?

creator.go
```diff
@@ -0,0 +1,28 @@
...
+package apmqueue
+
+import "context"
+
+// TopicCreator is an interface for creating topics.
+type TopicCreator interface {
+	// CreateTopics creates one or more topics.
+	//
+	// CreateTopics will leave existing topics untouched.
+	CreateTopics(ctx context.Context, topic ...Topic) error
```
dmathieu, line 27: Should `DeleteTopics` also be defined in an interface?

pubsublite/consumer_test.go
```diff
@@ -37,37 +37,3 @@ func TestNewConsumer(t *testing.T) {
 		assert.ErrorContains(t, err, "pubsublite: delivery is not valid")
 	})
 }
-
-func TestSubscriptionString(t *testing.T) {
```
dmathieu, line 41: Should there be tests for `Validate()`?
(maybe in another PR)

pubsublite/topiccreator.go
```diff
@@ -0,0 +1,164 @@
...
+var _ apmqueue.TopicCreator = (*TopicCreator)(nil)
+
+// TopicCreator creates GCP Pub/Sub topics.
+type TopicCreator struct {
+	m   *Manager
+	cfg TopicCreatorConfig
+}
+
+// NewTopicCreator returns a new TopicCreator with the given config.
+func (m *Manager) NewTopicCreator(cfg TopicCreatorConfig) (*TopicCreator, error) {
```
dmathieu, line 99: For kafka, the topic creation is held in the manager, while pubsublite has its own independent struct. What is the reason for not putting this logic within the `pubsublite.Manager` struct?

kafka/manager.go
```diff
@@ -0,0 +1,204 @@
...
+		return err
+	}
+	return nil
```
marclop: No need to check the error since we aren't wrapping it after it's been joined.
```suggestion
	return errors.Join(createErrors...)
```

kafka/manager.go
```diff
@@ -0,0 +1,204 @@
...
+	))
+	defer span.End()
+
+	topicNames := make([]string, len(topics))
+	for i, topic := range topics {
+		topicNames[i] = string(topic)
+	}
+	responses, err := m.client.DeleteTopics(ctx, topicNames...)
+	if err != nil {
+		return fmt.Errorf("failed to delete kafka topics: %w", err)
```
marclop, line 173:
```suggestion
		span.RecordError(err)
		return fmt.Errorf("failed to delete kafka topics: %w", err)
```

kafka/manager.go
```diff
@@ -0,0 +1,204 @@
...
+			}
+			continue
+		}
+		logger.Info("deleted kafka topic")
+	}
+	if err := errors.Join(deleteErrors...); err != nil {
+		return err
+	}
+	return nil
+
```
marclop:
```suggestion
	return errors.Join(deleteErrors...)
```

creator.go
axw: It's not needed at the moment. The code I intend to use it in will not be deleting topics, only creating.

I'll remove the `apmqueue.TopicCreator` interface for now, since it's not actually used anywhere in this PR.

pubsublite/consumer_test.go
axw: Yes probably, in a followup.

pubsublite/topiccreator.go
axw: This is so we can have a common CreateTopics signature/interface.

pubsublite has this concept of a throughput reservation, which kafka doesn't have. The pubsublite.Manager is not bound to a specific reservation (it can't be, it manages them); the TopicCreator is bound to one, so its CreateTopics method can create topics attached to that reservation.

dmathieu (approved):

marclop (approved):

pubsublite/consumer_test.go
axw: I expanded the tests and validation logic in this PR, added a commit.

kafka/manager.go
axw: fixed

kafka/manager.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	responses, err := m.client.CreateTopics(
+		ctx,
+		int32(m.cfg.PartitionCount),
+		-1, // default.replication.factor
+		m.topicConfigs,
+		topicNames...,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create kafka topics: %w", err)
```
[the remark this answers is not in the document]
axw: Good call, I'll push a commit to record & set status here and in the other spots

PR #170, opened 2023-05-25
Add system-test workflow

reakaleek:


.github/workflows/system-test.yml
```diff
@@ -0,0 +1,52 @@
+name: system-test
+
+on:
+  pull_request: ~
```
reakaleek, line 4: We have to remove this. Otherwise, this will always fail on fork PRs. (Because they don't have access to to secrets)

```suggestion
```

marclop (commented):
Pinging @axw given that #172 introduces the topic manager and it removes the need for the topic operator going forward. There's also a few changes in there that may be breaking (removal of terraform use for instance).

@axw Do you have any preference on merge order? The PR that gets merged last will need to make some changes and resolve conflicts.

axw:
@marclop thanks for the ping. Looks like this PR is ready, so I think let's merge it first and I'll adapt my PR.

marclop (approved):
LGTM code-wise. Thank you for making all the required changes in the Go code.

.github/workflows/system-test.yml
```diff
@@ -0,0 +1,52 @@
...
+  contents: read
+
+jobs:
+  run-system-tests:
+    timeout-minutes: 60
+    permissions:
+      id-token: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
amannocci, line 19:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/system-test.yml
reakaleek: I was think wrongly. We don't have access to secrets in a workflow from a forked PRs, not matter what.

We can only run this on main for now.

@marclop WDYT?

amannocci (changes requested):

.github/workflows/system-test.yml
reakaleek: https://github.com/elastic/apm-queue/pull/170/commits/2fde13409a8d39b14b510e3ea652d33f39d60b63 should fix the problem.

A workflow with the `pull_request_target` trigger always runs the workflow that is checked-in in the `main` branch.

Using the `is-pr-approved` action, we can check the conditions in [Motivation & Context]( https://github.com/elastic/apm-pipeline-library/tree/main/.github/actions/is-pr-approved#motivation--context).

only then the workflow is continued.

.github/workflows/system-test.yml
```diff
@@ -0,0 +1,66 @@
...
+        with:
+          terraform_version: 1.4.6
+      - uses: hashicorp/vault-action@v2.5.0
+        with:
+          url: ${{ secrets.VAULT_ADDR }}
+          roleId: ${{ secrets.VAULT_ROLE_ID }}
+          secretId: ${{ secrets.VAULT_SECRET_ID }}
+          method: approle
+          secrets: |
+            secret/observability-team/ci/service-account/apm-queue-system-tests credentials | GOOGLE_CREDS ;
```
kuisathaverat, line 44: This will not work in forks, it should be in another workflow triggered by `on_workflow`

reakaleek: > For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission unless the permissions key is specified and the workflow can access secrets, even when it is triggered from a fork.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

kuisathaverat: There is a Organization policy that do es not allow forks to access to secrets, so it will not work.

reakaleek: I had to verify this.

https://github.com/elastic/test-pull-request-target/actions/runs/5133114081/jobs/9235276648?pr=2

The secret is hidden with `***` but it seems to work.

kuisathaverat (approved):

kuisathaverat (approved):

amannocci (approved):
LGTM

marclop (approved):
LGTM! TIL about `is-pr-approved`.

.github/workflows/system-test.yml
marclop: Awesome!

reakaleek:
Merging: Failing tests are unrelated to the changes and are currently also failing on the main branch.

Issue #95, opened 2023-04-12
system tests: Run as part of PR checks


marclop:
## Description

We MUST ensure that merged PRs do not break the system tests, and that these are run for each PR and are required to pass before PRs are merged.

Since the system tests will provision external infrastructure, only PRs opened by approved contributors will run these automatically.

## Related issues

Requires #92

simitt:
@elastic/observablt-robots can you take this task on?

cachedout:
@simitt Yes of course. I've added this to the iteration taking place between May 18 and June 7 which should give you plenty of headroom for 8.9.

@v1v Could you find an owner for this please?

v1v:
> @v1v Could you find an owner for this please?

Sure, @reakaleek is assigned. We have planned this task to happen in this upcoming iteration (`Iteration 22` which starts in 2 days and last for 3 weeks)

reakaleek:
Cool. Thank you!

reakaleek:
Hey @marclop 

I just tried this with 

```shell
go version
go version go1.20.4 darwin/arm64
```

which results in:

```shell
go run -v -timeout=30m .
flag provided but not defined: -timeout
usage: go run [build flags] [-exec xprog] package [arguments...]
Run 'go help run' for details.
```

when I omit the `-timeout` argument, I get this:

```
go run -v .             
package github.com/elastic/apm-queue/systemtest is not a main package
```

Sorry, I'm not too familiar with go. No clue on how to make this run.

marclop:
@reakaleek Oops, sorry. I miss typed. The right command is `go test`.

reakaleek:
@marclop I've got to a point where the `helm_release` resources are installed correctly by terraform.

however, the `kafkatopic` is never becomes `Ready=True` and then the command times out.

```
k get kafkatopics
NAME                                                                                               CLUSTER   PARTITIONS   REPLICATION FACTOR   READY
consumer-offsets---84e7a678d08f4bd226872e5cdd4eb527fadc1c6a                                        kafka     50           1                    True
dummy.gdvjnzso                                                                                     kafka     1            1                    
strimzi-store-topic---effb8e3e057afce1ecf67c3f5d8e4e3ff177fc55                                     kafka     1            1                    True
strimzi-topic-operator-kstreams-topic-store-changelog---b75e702040b99be8a9263134de3507fc0cc4017b   kafka     1            1                    True
```

do you maybe have an idea why this is the behaviour?

[...]

PR #172, opened 2023-05-25
systemtest: use manager types to create topics

axw:
Rather than creating topics through Terraform, and Strimzi, create them using managers. We now provision the cluster during initialisation (main_test.go), and each test is responsible for creating (and destroying) its topics.

Port forwarding is now done per test. I found that the port forwarder would eventually get stuck and cause tests to time out -- I think something is not closing connections, but I couldn't get to the bottom of it.

`TestShutdown/Kafka/close` is failing, presumably due to the deadlock that should be fixed by https://github.com/elastic/apm-queue/pull/173 (I haven't confirmed)


systemtest/infra_kafka.go
```diff
@@ -21,233 +21,203 @@ import (
...
+// deleting them when the test completes.
+func CreateKafkaTopics(ctx context.Context, t testing.TB, topics ...apmqueue.Topic) {
+	manager, err := NewKafkaManager(t)
+	require.NoError(t, err)
+	defer manager.Close()
+	err = manager.CreateTopics(ctx, topics...)
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		manager, err := NewKafkaManager(t)
```
marclop, line 126: Would it make sense to just create a single kafka manage3r and only close it after the topics are deleted in the `t.Cleanup` function?

systemtest/infra_kafka.go
```diff
@@ -21,233 +21,203 @@ import (
...
+func KafkaCommonConfig(t testing.TB) kafka.CommonConfig {
+	commonConfig := kafka.CommonConfig{
+		Brokers: append([]string{}, kafkaBrokers...),
+		Logger:  logger().Desugar().Named("kafka"),
+	}
+	if len(commonConfig.Brokers) == 0 {
+		brokerAddress := portforwardKafka(t)
+		tlsDialer := &tls.Dialer{
+			// running locally, should be fast
+			NetDialer: &net.Dialer{Timeout: time.Second},
```
marclop, line 210: we may need to increase this after #95

systemtest/infra_kafka.go
```diff
@@ -21,233 +21,203 @@ import (
...
+			Config:    &tls.Config{InsecureSkipVerify: true},
+		}
+		commonConfig.Brokers = []string{brokerAddress}
+		commonConfig.Dialer = func(ctx context.Context, network, addr string) (net.Conn, error) {
+			// The advertised broker address is only reachable within
+			// the Kubernetes cluster; replace it with the localhost
+			// port-forwarded address.
+			addr = brokerAddress
```
marclop, line 218: ah nice, so this removes the need for the mockdns?

marclop (approved):

systemtest/infra_kafka.go
```diff
@@ -21,233 +21,203 @@ import (
...
-				cfg.Name,
-				cfg.PortForwardResource,
-				cfg.PortForwardMapping,
-			)
-		})
-		if err != nil {
-			return fmt.Errorf("failed to run port forward: %w", err)
-		}
+
+	// Install Strimzi using Helm.
```
reakaleek, line 78: There are some complications in installing the strimzi helm chart multiple times in the same cluster when multiple PRs cause concurrent installations of strimzi.

There is a conflict with ClusterRoles and ClusterRole bindings.

reakaleek: In https://github.com/elastic/apm-queue/pull/170 I avoided this, by removing the strimzi part and installing a permantent strimzi operator that listens on any namespace.

systemtest/infra_kafka.go
axw: Yes, that would be better. I'll create two cleanups: one to close, one to delete topics.

systemtest/infra_kafka.go
axw: yes, there's only one broker, so we can just ignore the advertised address

systemtest/infra_kafka.go
axw: Thanks, I've remove the Strimzi installation from this PR. Longer term I think we should probably stop relying on Strimzi, and try to use a single Kafka broker pod with ephemeral storage.

systemtest/infra_kafka.go
axw: It's working fine at the moment, but let's keep an eye on it and update if needed. I'd update it now but I don't want to wait another 15-20 minutes to merge :smile:

PR #173, opened 2023-05-25
consumer(kafka): Fix close deadlock, update errors

### LINKED ISSUE #147: consistent behaviour on shutdown
The two providers have different behaviours on shutdown 

Add a test to ensure the behaviour is consistent

### LINKED ISSUE #157: kafka: deadlock when closing consumer without canceling context
kafka consumer Run method will block forever unless the context is canceled.

Trying to close the consumer with `consumer.Close` will block forever. 

This only happens with the Kafka consumer. Pubsublite is not affected.

To reproduce: `cd systemtest && go test -v -run=TestShutdown/Kafka/close`

consumer will block at: https://github.com/elastic/apm-queue/blob/e48ac80251203194fe79626b14749455d5f290d8/kafka/consumer.go#L189

The close method will block trying to acquire the lock.

marclop:
## Description
Updates the `kafka.Consumer` to behave like the `pubsublite.Consumer`. This includes:

- Returning a `nil` error after the context is cancelled, or the client is closed.
- Returning an error if the consumer has already been started.

## Related issues

Closes #157 
Closes #147

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 
 	"github.com/twmb/franz-go/pkg/kgo"
 	"go.opentelemetry.io/otel/trace"
@@ -100,8 +101,8 @@ type Consumer struct {
 	client   *kgo.Client
 	cfg      ConsumerConfig
 	consumer *consumer
-	running  chan struct{}
-	closed   chan struct{}
+	running  atomic.Bool
+	closed   atomic.Bool
 
 	tracer trace.Tracer
 }
@@ -151,23 +152,19 @@ func NewConsumer(cfg ConsumerConfig) (*Consumer, error) {
 		cfg:      cfg,
 		client:   client,
 		consumer: consumer,
-		closed:   make(chan struct{}),
-		running:  make(chan struct{}),
 		tracer:   cfg.tracerProvider().Tracer("kafka"),
 	}, nil
 }
 
 // Close the consumer, blocking until all partition consumers are stopped.
 func (c *Consumer) Close() error {
-	select {
-	case <-c.closed:
-	default:
-		c.mu.Lock()
-		defer c.mu.Unlock()
-		c.client.Close()
-		c.consumer.wg.Wait() // Wait for all the goroutines to exit.
-		close(c.closed)
+	if !c.closed.CompareAndSwap(false, true) {
+		return nil
 	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.client.Close()
+	c.consumer.wg.Wait() // Wait for all the goroutines to exit.
 	return nil
 }
 
@@ -177,11 +174,8 @@ func (c *Consumer) Close() error {
 // To shut down the consumer, cancel the context, or call consumer.Close().
 // If called more than once, returns `apmqueue.ErrConsumerAlreadyRunning`.
 func (c *Consumer) Run(ctx context.Context) error {
-	select {
-	case <-c.running:
+	if !c.running.CompareAndSwap(false, true) {
 		return apmqueue.ErrConsumerAlreadyRunning
-	default:
-		close(c.running)
 	}
 	for {
 		if err := c.fetch(ctx); err != nil {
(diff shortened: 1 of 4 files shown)


kafka/consumer.go
```diff
@@ -182,20 +197,18 @@ func (c *Consumer) Run(ctx context.Context) error {
...
-	}
-	if errors.Is(fetches.Err0(), context.Canceled) ||
+	if fetches.IsClientClosed() ||
+		errors.Is(fetches.Err0(), context.Canceled) ||
 		errors.Is(fetches.Err0(), context.DeadlineExceeded) {
-		return fmt.Errorf("context done: %w", fetches.Err0())
+		return context.Canceled
 	}
+	// Acquire the lock after checking if the the context is cancelled and/or
+	// the client is closed.
```
kruskall, line 209: **issue**: I don't think we can do this. What happens if the context is canceled or the client is closed before we acquire the lock but after the records were polled ? This is also making the previous if not safe for concurrent usage as the client could be closed after the `IsClientClosed` is called.

kruskall (changes requested):

kafka/consumer.go
marclop: > I don't think we can do this. What happens if the context is canceled or the client is closed before we acquire the lock but after the records were polled ? 

If the context is cancelled before the lock is acquired, the call on line 217 will fail for `AtMostOnceDelivery`, and the function will return then. No records will be lost since the commit for the fetched offsets will have failed.

If the context is cancelled between line 228 and 240, the records will be lost, since `AtMostOnceDelivery` has that caveat. This is no different with the lock acquired at the top of the function, before `PollRecords` is called, or after.

The locking doesn't prevent the context from being cancelled at any point in the execution flow.

> This is also making the previous if not safe for concurrent usage as the client could be closed after the IsClientClosed is called.

Why would closing the `kgo.Client` make calls to it not safe? The client can be called at any time, and the `kgo.Client` is safe for concurrent use. The client can be closed at any time while other methods are called, and is within the intended use scope.

kafka/consumer.go
```diff
@@ -149,31 +151,44 @@ func NewConsumer(cfg ConsumerConfig) (*Consumer, error) {
...
 //
 // To shut down the consumer, cancel the context, or call consumer.Close().
+// If called more than once, returns `apmqueue.ErrConsumerAlreadyRunning`.
 func (c *Consumer) Run(ctx context.Context) error {
+	select {
+	case <-c.running:
+		return apmqueue.ErrConsumerAlreadyRunning
+	default:
+		close(c.running)
+	}
```
lahsivjar, line 185: This can lead to panic due to multiple close on a channel if `Run` is called multiple times in quick succession. As an alternative what do you think about using `atomic.Bool` with its `CompareAndSwap` operation:

```suggestion
        // c.running is defined as an atomic.Bool
	if !c.running.CompareAndSwap(false, true) {
	        return apmqueue.ErrConsumerAlreadyRunning
	}
```

kafka/consumer.go
```diff
@@ -149,31 +151,44 @@ func NewConsumer(cfg ConsumerConfig) (*Consumer, error) {
...
 // Close the consumer, blocking until all partition consumers are stopped.
 func (c *Consumer) Close() error {
-	c.mu.Lock()
-	defer c.mu.Unlock()
-	// Since we take the full lock when closing, there's no need to explicitly
-	// allow rebalances since polls aren't concurrent with Close().
-	c.client.Close()
-	c.consumer.wg.Wait() // Wait for all the goroutines to exit.
+	select {
+	case <-c.closed:
```
lahsivjar, line 163: For `c.closed` also we can use an `atomic.Bool` as for `c.running`, however, there is no panic in this case.

kafka/consumer.go
kruskall: TIL! Thanks for the explanation! bow

My only other concern would be that the consumer is not guaranteed to return nil on shutdown. For example, in one of the case you mentioned:

    If the context is cancelled before the lock is acquired, the call on line 217 will fail for AtMostOnceDelivery, and the function will return then.

the consumer would return ErrCommitFailed. WDYT ?

marclop: I think that's a legitimate error, since in that case, we may want to re-create the consumer, rather than an entire process.

kruskall (approved):

kafka/consumer.go
marclop: Fixed in 9308bfe

lahsivjar (approved):

PR #174, opened 2023-05-26
pubsublite: extract project_id from credentials

axw:
Service account credentials have an associated project, use that as the default for CommonConfig.Project.

### DIFF
--- a/pubsublite/common.go
+++ b/pubsublite/common.go
@@ -18,7 +18,10 @@
 package pubsublite
 
 import (
+	"encoding/json"
 	"errors"
+	"fmt"
+	"os"
 
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/trace"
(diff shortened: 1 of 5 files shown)


pubsublite/common.go
```diff
@@ -64,6 +70,45 @@ func (cfg CommonConfig) Validate() error {
 	return errors.Join(errs...)
 }
 
+// setFromEnv sets unspecified config from environment variables.
+//
+// If $GOOGLE_APPLICATION_CREDENTIALS is set, the file to which it points will
+// be read and parsed as JSON (assuming service account), and its `project_id`
+// property will be used for the Project property if it is not already set.
+func (c *CommonConfig) setFromEnv() error {
```
dmathieu, line 78: (to fix lint issue)

```suggestion
func (cfg *CommonConfig) setFromEnv() error {
```

dmathieu (approved):

pubsublite/common.go
axw: thanks!

PR #175, opened 2023-05-26
consumer(kafka): Use record.Context in processor

marclop:
## Description
Use each of the record's context in `processor.ProcessBatch` in order to
propagate the record's headers in the context. Update the `consumer` to
implement the `OnFetchRecordBuffered` hook and set the record.Context if
nil.

This allows the `kgo.Client.Close()` call to cancel the first context and trigger shutdown to stop record processing.

## Context

The distributed traces were broken without this fix.


kafka/consumer.go
```diff
@@ -352,7 +352,10 @@ func (pc partitionConsumer) consume(ctx context.Context, topic string, partition
 				// NOTE(marclop) The decoding has failed, a DLQ may be helpful.
 				continue
 			}
-			ctx := queuecontext.WithMetadata(ctx, meta)
+			// NOTE(marclop) Using the kgo.Client context may result in records
+			// not being processed by the configured process if the client is
+			// closed while some records are processed.
+			ctx := queuecontext.WithMetadata(combineContexts(ctx, msg.Context), meta)
```
axw, line 358: kotel will initialise msg.Context with context.Background, and add the trace context to it:

https://github.com/twmb/franz-go/blob/8406043eb1a3eb5e117336b373106a5a9003c9f6/plugin/kotel/tracer.go#L229-L231

Should we add our own hook before motel that sets the base context to one that is cancelled when the client is closed? Then just use msg.Context here directly?

axw (approved):

kafka/consumer.go
marclop: Good point. I'll make the change rather than using the `combineContexts` function.

marclop: fixed in f22f47a

PR #178, opened 2023-05-29
consumer(kafka): Fix consumer graceful shutdown

### LINKED ISSUE #161: test: TestConsumerDelivery is flaky
`TestConsumerDelivery` is failing inconsistently

Investigate what's wrong

marclop:
## Description

Fixes the consumer graceful shutdown by closing the partition consumers
first, and only after they've all been stopped, close the kgo.Client to
allow the records to be processed before the client is closed.

This fixes some flaky tests and inconsistent shutdown behavior.

## Related issues

Closes #161


axw (commented):
The test is still failing for me when I run it with `-race` in a loop:

`go test -race -run TestConsumerGracefulShutdown -failfast -count=1000`

What is the purpose of this test? It's not clear to me what "graceful shutdown" means, looking at what the test does. Given that we're closing the kgo.Client before the events are known to be consumed, why would we expect them all to be consumed?

kafka/consumer.go
```diff
@@ -165,9 +165,9 @@ func (c *Consumer) Close() error {
 	if !c.closed.CompareAndSwap(false, true) {
```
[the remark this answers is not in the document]
lahsivjar: FYI: It was introduced in [this PR](https://github.com/elastic/apm-queue/pull/173) with the expectation that we will return nil if the client is already closed. Maybe what we need is `defer c.closed.Swap(true)`(??)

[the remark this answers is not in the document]
axw: :ok_hand:  I think we should generally avoid atomics -- channels are typically more appropriate when trying to communicate between goroutines, to avoid race conditions. If we need an atomic, then we can indeed `defer c.closed.Store(true)`.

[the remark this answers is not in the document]
lahsivjar: > Maybe what we need is defer c.closed.Swap(true) if we just want to know if Close succeeded or not

I _think_ this in itself will not be enough, sorry for the confusion. The reason is that after checking with `c.closed.Load()` and before acquiring the read lock, the method `Close` can be called. This I assume will cause `fetch` to fail with a different error than expected `context.Canceled`. I think @axw's suggestion to use channels in this case and interrupt using the channel inside the read lock is better.

marclop:
I've made changes in the consumer to incorporate your feedback, and also fix the not so graceful `consumer.Close()` to wait for the partition consumers to exit.

endorama (commented):
I've reviewed the code, I've little context to judge the logic. Tests (with `-race`) are passing.

kafka/consumer.go
```diff
@@ -203,9 +227,16 @@ func (c *Consumer) fetch(ctx context.Context) error {
 		errors.Is(fetches.Err0(), context.DeadlineExceeded) {
 		return context.Canceled
 	}
-	// Acquire the lock after checking if the the context is cancelled and/or
-	// the client is closed.
-	c.mu.RLock()
+	// Acquire the lock after checking if the the context is done.
+	// If the lock cannot be acquired, the consumer is closing,
+	// be closing, in which case, return with context.Cancelled.
+	if !c.mu.TryRLock() {
```
lahsivjar, line 233: Not a fan of `TryRLock` usage. WDYT about interrupting the `for` loop in `Run` if `c.closed` is closed and in the `Close` method, waiting for `Run` to finish before proceeding to close stuff. Example (only for conveying the intention):

```go
func (c *Consumer) Run(ctx context.Context) error {
	c.mu.Lock()
	select {
	case <-c.running:
		c.mu.Unlock()
		return apmqueue.ErrConsumerAlreadyRunning
	default:
		close(c.running)
	}
	c.mu.Unlock()
        defer close(c.runStopped)
	for {
                select {
                case <-c.closed:
                    return nil // or ErrClosed
                default:
                }
		if err := c.fetch(ctx); err != nil {
			if errors.Is(err, context.Canceled) {
				return nil // Return no error if err == context.Canceled.
			}
			return err
		}
	}
}
```

Then in `Close`:

```go
	c.mu.Lock()
	select {
	case <-c.closed:
	default:
		close(c.closed)
        }
        c.mu.Unlock()
        <-c.runStopped // can choose to pass a context to Close and wait with select

        c.mu.Lock()
        defer c.mu.Unlock()
        // Close all partition consumers first to ensure that there aren't
        // any records being processed while the kgo.Client is being closed.
        // Also ensures that commits can be issued after the records are
        // processed when AtLeastOnceDelivery is configured.
        c.consumer.close() // if this requires to be done only once, we can use a bool flag or maybe set the values to nil if successfully closed
        c.client.Close()
	c.consumer.wg.Wait() // Wait for consumer goroutines to exit.
	return nil
}
```

kafka/consumer.go
```diff
@@ -203,9 +227,16 @@ func (c *Consumer) fetch(ctx context.Context) error {
 		errors.Is(fetches.Err0(), context.DeadlineExceeded) {
 		return context.Canceled
 	}
-	// Acquire the lock after checking if the the context is cancelled and/or
-	// the client is closed.
-	c.mu.RLock()
+	// Acquire the lock after checking if the the context is done.
+	// If the lock cannot be acquired, the consumer is closing,
+	// be closing, in which case, return with context.Cancelled.
```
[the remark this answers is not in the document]
marclop: yep, that's correct. I botched this comment a bit. I'll fix it.

marclop:
@lahsivjar @axw I've pushed 1fd36a0, which uses a more solid approach suggested by Andrew. Let me know what you think about it. After this PR is merged, calling `consumer.Close()` will trigger graceful shutdown and either successfully shutdown all the partition consumers within `cfg.ShutdownGracePeriod`, or trigger forceful shutdown after that duration is exceeded.

kafka/consumer.go
```diff
@@ -153,39 +174,80 @@ func NewConsumer(cfg ConsumerConfig) (*Consumer, error) {
...
+		c.mu.Unlock()
 		return apmqueue.ErrConsumerAlreadyRunning
+	default:
+		close(c.running)
 	}
+	// Create a new context from the passed context, used exclusively for
+	// kgo.Client.* calls. c.stopFetch is called by consumer.Close() to
+	// cancel this context as part of the graceful shutdown sequence.
+	var clientCtx context.Context
+	clientCtx, c.stopPoll = context.WithCancel(ctx)
```
axw, line 245: Because `stopPoll` is only set after `Run` is called, calling `Stop` before `Run` means that the poll context won't be closed. Should we instead start a goroutine that waits for `c.closed` and cancels the context?

kafka/consumer.go
```diff
@@ -153,39 +174,80 @@ func NewConsumer(cfg ConsumerConfig) (*Consumer, error) {
...
+		// the underlying processing context if they aren't stopped in time.
+		select {
+		case <-time.After(c.cfg.ShutdownGracePeriod): // Timeout
+			c.forceClose(fmt.Errorf(
+				"consumer: close: timeout waiting for consumers to stop (%s)",
+				c.cfg.ShutdownGracePeriod.String(),
+			))
+		case <-stopped: // Stopped within c.cfg.ShutdownGracePeriod.
+		}
+		c.client.Close() // Last, close the `kgo.Client`
```
axw, line 218: Should we do this as a defer, after the `close(c.closed)`? Or does it need to happen before we wait for the waitgroup?

kafka/consumer.go
```diff
@@ -153,39 +174,80 @@ func NewConsumer(cfg ConsumerConfig) (*Consumer, error) {
...
+	}
+	c.consumer.wg.Wait() // Wait for consumer goroutines to exit.
```
axw, line 220: Maybe do this in a defer, and then we can early return if we hit `case <-c.closed`

kafka/consumer.go
```diff
@@ -322,23 +353,97 @@ func (c *consumer) assigned(ctx context.Context, client *kgo.Client, assigned ma
...
+		case <-ctx.Done(): // Consumer.Run(ctx) context.
+			err = ctx.Err()
+		case <-c.ctx.Done(): // Forceful cancellation context.
+			err = c.ctx.Err()
+		}
+		// When using AtMostOnceDelivery, if the context is cancelled between
+		// PollRecords and this line, records will be lost.
+		if c.delivery == apmqueue.AtMostOnceDeliveryType {
+			c.logWarn(err, ftp)
+		}
```
axw, line 427:
```suggestion
		if err != nil {
			if c.delivery == apmqueue.AtMostOnceDeliveryType {
				c.logWarn(err, ftp)
			}
		}
```

kafka/consumer.go
```diff
@@ -322,23 +353,97 @@ func (c *consumer) assigned(ctx context.Context, client *kgo.Client, assigned ma
...
+}
+
 // Implement the kgo.Hook that allows injecting the kgo.Client context that is
 // received on partition assignment.
 func (c *consumer) OnFetchRecordBuffered(r *kgo.Record) {
 	// Inject the context so records's context have cancellation.
 	if r.Context == nil {
-		c.mu.RLock()
-		defer c.mu.RUnlock()
 		// NOTE(marclop) Using the kgo.Client context may result in records
```
axw, line 447: We're not using it now, right? Should we update this comment?

kafka/consumer_test.go
```diff
@@ -421,26 +438,44 @@ func TestConsumerGracefulShutdown(t *testing.T) {
...
+		// Assert the consumer error.
+		switch dt {
+		case apmqueue.AtMostOnceDeliveryType:
+			if consumerErr != nil {
+				assert.Equal(t, ErrCommitFailed, consumerErr)
+			}
+		case apmqueue.AtLeastOnceDeliveryType:
+			assert.NoError(t, consumerErr)
+		}
+		assert.GreaterOrEqual(t, processed.Load(), int32(1)) //  Ensure that at least one record is processed.
```
axw, line 478: Isn't the objective of the test to ensure both records are processed?

kafka/consumer_test.go
```diff
@@ -421,26 +438,44 @@ func TestConsumerGracefulShutdown(t *testing.T) {
...
+		case <-time.After(time.Second):
+			t.Fatal("timed out waiting for consumer to exit")
+		}
+
+		// Assert the consumer error.
+		switch dt {
+		case apmqueue.AtMostOnceDeliveryType:
+			if consumerErr != nil {
+				assert.Equal(t, ErrCommitFailed, consumerErr)
+			}
```
axw, line 474: Should this error be consistent/deterministic?

axw (commented):
Thanks, I think this is looking good.

kafka/consumer.go
marclop: I suppose it can be done as a defer, but it seemed more appropriate to keep it outside of the defer just to be able to visualize the order of calls more easily. I can add it in a defer after `close(c.closed)` if you think that's better.

kafka/consumer.go
marclop: sorry, but I'm afraid I don't follow. The `clientCtx` is only used if `Run()` is called, otherwise it's not used. `c.stopPoll` is set to an empty function and on close would be a no-op.

axw:
```go
consumer.Close()
consumer.Run() <- clientCtx is not cancelled because Run was called after Close returned
```

I suppose it doesn't matter though, because Close will close the kgo.Client, causing fetch to return context.Canceled, breaking the loop.

kafka/consumer.go
axw: I'm mildly concerned that we might update the code and introduce a code path that would lead to the kgo.Client being left open. I'll leave it to your judgement, just a passing thought.

kafka/consumer_test.go
marclop: fixed!

kafka/consumer_test.go
marclop: correct, fixed.

kafka/consumer.go
marclop: done

axw (approved):

PR #181, opened 2023-05-30
Enhance kgo metrics for producer

endorama:
This PR adds additional metrics to `kafka` package producer to enhance the one provided by `kotel`.

It adds support for a custom meter (instead of always relying on the global one) and provides additional metric by leveraging `kgo` hooks.

Additional metrics provided: 
- `producer.messages.produced`
- `producer.messages.errors`

~While testing this I noticed the provided test has some flakiness, due to the fact that the code instantiate 2 different meters and it may happen the order of metrics in `rm.ScopeMetrics` is different. As the test is mainly useful for development, I moved it to a separate file under the `metrics` build tag to prevent it affect other tests.~

### DIFF
--- a/kafka/common.go
+++ b/kafka/common.go
@@ -194,7 +194,7 @@ func (cfg *CommonConfig) newClient(additionalOpts ...kgo.Opt) (*kgo.Client, erro
 		)
 		metricHooks, err := NewKgoHooks(cfg.meterProvider())
 		if err != nil {
-			return nil, fmt.Errorf("cannot setup kgo metrics hooks: %w", err)
+			return nil, fmt.Errorf("kafka: failed creating kgo metrics hooks: %w", err)
 		}
 		opts = append(opts,
 			kgo.WithHooks(metricHooks),
--- a/kafka/common.go
+++ b/kafka/common.go
@@ -192,8 +192,12 @@ func (cfg *CommonConfig) newClient(additionalOpts ...kgo.Opt) (*kgo.Client, erro
 			kotel.WithTracer(kotel.NewTracer(kotel.TracerProvider(cfg.tracerProvider()))),
 			kotel.WithMeter(kotel.NewMeter(kotel.MeterProvider(cfg.meterProvider()))),
 		)
+		metricHooks, err := NewKgoHooks(cfg.meterProvider())
+		if err != nil {
+			return nil, fmt.Errorf("cannot setup kgo metrics hooks: %w", err)
+		}
 		opts = append(opts,
-			kgo.WithHooks(NewKgoHooks(cfg.meterProvider())),
+			kgo.WithHooks(metricHooks),
 			kgo.WithHooks(kotelService.Hooks()...),
 		)
 	}
--- a/kafka/common.go
+++ b/kafka/common.go
@@ -192,7 +192,10 @@ func (cfg *CommonConfig) newClient(additionalOpts ...kgo.Opt) (*kgo.Client, erro
 			kotel.WithTracer(kotel.NewTracer(kotel.TracerProvider(cfg.tracerProvider()))),
 			kotel.WithMeter(kotel.NewMeter(kotel.MeterProvider(cfg.meterProvider()))),
 		)
-		opts = append(opts, kgo.WithHooks(kotelService.Hooks()...))
+		opts = append(opts,
+			kgo.WithHooks(NewKgoHooks(cfg.meterProvider())),
+			kgo.WithHooks(kotelService.Hooks()...),
+		)
 	}
 	client, err := kgo.NewClient(opts...)
 	if err != nil {
--- a/kafka/common.go
+++ b/kafka/common.go
@@ -192,7 +192,7 @@ func (cfg *CommonConfig) newClient(additionalOpts ...kgo.Opt) (*kgo.Client, erro
 			kotel.WithTracer(kotel.NewTracer(kotel.TracerProvider(cfg.tracerProvider()))),
 			kotel.WithMeter(kotel.NewMeter(kotel.MeterProvider(cfg.meterProvider()))),
 		)
-		metricHooks, err := NewKgoHooks(cfg.meterProvider())
+		metricHooks, err := newKgoHooks(cfg.meterProvider())
 		if err != nil {
 			return nil, fmt.Errorf("kafka: failed creating kgo metrics hooks: %w", err)
 		}
--- a/kafka/metrics.go
+++ b/kafka/metrics.go
@@ -1,3 +1,20 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
 package kafka
 
 import (
--- a/kafka/metrics.go
+++ b/kafka/metrics.go
@@ -76,30 +76,25 @@ func newKgoHooks(mp metric.MeterProvider) (*kgoHooks, error) {
 
 // https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#HookProduceRecordUnbuffered
 func (h *kgoHooks) OnProduceRecordUnbuffered(r *kgo.Record, err error) {
-	partitionDimension := attribute.Int("partition", int(r.Partition))
-	topicDimension := attribute.String("topic", r.Topic)
+	partition := attribute.Int("partition", int(r.Partition))
+	topic := attribute.String("topic", r.Topic)
 
 	if err != nil {
-		errorDimension := attribute.String("error", "other")
-
+		errorType := attribute.String("error", "other")
 		if errors.Is(err, context.DeadlineExceeded) {
-			errorDimension = attribute.String("error", "timeout")
+			errorType = attribute.String("error", "timeout")
 		}
 		if errors.Is(err, context.Canceled) {
-			errorDimension = attribute.String("error", "canceled")
+			errorType = attribute.String("error", "canceled")
 		}
 
-		h.instruments.writeErrors.Add(
-			context.Background(),
-			1,
-			metric.WithAttributes(partitionDimension, topicDimension, errorDimension),
+		h.instruments.writeErrors.Add(context.Background(), 1,
+			metric.WithAttributes(partition, topic, errorType),
 		)
 		return
 	}
 
-	h.instruments.messageProduced.Add(
-		context.Background(),
-		1,
-		metric.WithAttributes(partitionDimension, topicDimension))
+	h.instruments.messageProduced.Add(context.Background(), 1,
+		metric.WithAttributes(partition, topic))
 
 }
--- a/kafka/metrics.go
+++ b/kafka/metrics.go
@@ -25,6 +25,7 @@ import (
 	"github.com/twmb/franz-go/pkg/kgo"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
+	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
 )
 
 const (
@@ -76,7 +77,7 @@ func newKgoHooks(mp metric.MeterProvider) (*kgoHooks, error) {
 
 // https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#HookProduceRecordUnbuffered
 func (h *kgoHooks) OnProduceRecordUnbuffered(r *kgo.Record, err error) {
-	partition := attribute.Int("partition", int(r.Partition))
+	partition := semconv.MessagingKafkaDestinationPartition(int(r.Partition))
 	topic := attribute.String("topic", r.Topic)
 
 	if err != nil {
--- a/kafka/metrics.go
+++ b/kafka/metrics.go
@@ -2,7 +2,7 @@ package kafka
 
 import (
 	"context"
-	"runtime"
+	"errors"
 	"time"
 
 	"github.com/twmb/franz-go/pkg/kgo"
@@ -62,6 +62,36 @@ func NewKgoHooks(mp metric.MeterProvider) *kgoHooks {
 	}
 }
 
+func (h *kgoHooks) OnProduceRecordUnbuffered(r *kgo.Record, err error) {
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			h.instruments.WriteTimeout.Add(
+				context.Background(),
+				1,
+				metric.WithAttributes(
+					attribute.Int("partition", int(r.Partition)),
+					attribute.String("topic", r.Topic),
+				))
+		}
+		h.instruments.WriteErrors.Add(
+			context.Background(),
+			1,
+			metric.WithAttributes(
+				attribute.Int("partition", int(r.Partition)),
+				attribute.String("topic", r.Topic),
+			))
+		return
+	}
+
+	h.instruments.MessageProduced.Add(
+		context.Background(),
+		1,
+		metric.WithAttributes(
+			attribute.Int("partition", int(r.Partition)),
+			attribute.String("topic", r.Topic),
+		))
+
+}
 
 // https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#HookBrokerWrite
 func (h *kgoHooks) OnBrokerWrite(meta kgo.BrokerMetadata, _ int16, _ int, writeWait, _ time.Duration, err error) {
--- a/kafka/metrics.go
+++ b/kafka/metrics.go
@@ -39,9 +39,9 @@ const (
 )
 
 type instruments struct {
-	MessageProduced metric.Int64Counter
-	WriteErrors     metric.Int64Counter
-	WriteTimeout    metric.Int64Counter
+	messageProduced metric.Int64Counter
+	writeErrors     metric.Int64Counter
+	writeTimeout    metric.Int64Counter
 }
 
 type kgoHooks struct {
@@ -80,9 +80,9 @@ func NewKgoHooks(mp metric.MeterProvider) (*kgoHooks, error) {
 
 	return &kgoHooks{
 		instruments{
-			MessageProduced: messageProducedCounter,
-			WriteErrors:     writeErrorCounter,
-			WriteTimeout:    writeTimeoutCounter,
+			messageProduced: messageProducedCounter,
+			writeErrors:     writeErrorCounter,
+			writeTimeout:    writeTimeoutCounter,
 		},
 	}, nil
 }
@@ -94,19 +94,19 @@ func (h *kgoHooks) OnProduceRecordUnbuffered(r *kgo.Record, err error) {
 	)
 	if err != nil {
 		if errors.Is(err, context.DeadlineExceeded) {
-			h.instruments.WriteTimeout.Add(
+			h.instruments.writeTimeout.Add(
 				context.Background(),
 				1,
 				metric.WithAttributeSet(attrs))
 		}
-		h.instruments.WriteErrors.Add(
+		h.instruments.writeErrors.Add(
 			context.Background(),
 			1,
 			metric.WithAttributeSet(attrs))
 		return
 	}
 
-	h.instruments.MessageProduced.Add(
+	h.instruments.messageProduced.Add(
 		context.Background(),
 		1,
 		metric.WithAttributeSet(attrs))
@@ -116,7 +116,7 @@ func (h *kgoHooks) OnProduceRecordUnbuffered(r *kgo.Record, err error) {
 // https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#HookBrokerWrite
 func (h *kgoHooks) OnBrokerWrite(meta kgo.BrokerMetadata, _ int16, _ int, writeWait, _ time.Duration, err error) {
 	if err != nil {
-		h.instruments.WriteErrors.Add(
+		h.instruments.writeErrors.Add(
 			context.Background(),
 			1,
 			metric.WithAttributes(
--- a/kafka/metrics.go
+++ b/kafka/metrics.go
@@ -16,6 +16,7 @@ const unitCount = "1"
 type instruments struct {
 	MessageProduced metric.Int64Counter
 	WriteErrors     metric.Int64Counter
+	WriteTimeout    metric.Int64Counter
 }
 
 type kgoHooks struct {
@@ -43,10 +44,20 @@ func NewKgoHooks(mp metric.MeterProvider) *kgoHooks {
 		panic(err)
 	}
 
+	c, err := m.Int64Counter(
+		"write.timeout.count",
+		metric.WithDescription("The total number of messages not produced due to timeout"),
+		metric.WithUnit(unitCount),
+	)
+	if err != nil {
+		panic(err)
+	}
+
 	return &kgoHooks{
 		instruments{
 			MessageProduced: a,
 			WriteErrors:     b,
+			WriteTimeout:    c,
 		},
 	}
 }
--- a/kafka/metrics.go
+++ b/kafka/metrics.go
@@ -35,13 +35,11 @@ const (
 
 	messageProducedCounterKey = "message.produced"
 	writeErrorCounterKey      = "write.error"
-	writeTimeoutCounterKey    = "write.timeout"
 )
 
 type instruments struct {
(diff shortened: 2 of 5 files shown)


kafka/metrics.go
```diff
@@ -0,0 +1,103 @@
...
+		instruments{
+			MessageBuffered: c,
+			MessageProduced: a,
+			WriteErrors:     b,
+		},
+	}
+}
+
+// https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#HookProduceBatchWritten
+func (h *kgoHooks) OnProduceBatchWritten(meta kgo.BrokerMetadata, topic string, partition int32, metrics kgo.ProduceBatchMetrics) {
```
marclop, line 66: Shouldn't this be `OnProduceRecordUnbuffered`? Otherwise any bytes written to the broker will increment the counter. It is implemented by the `kotel.Tracer` to end a span, but in this case, we could use it to record that a record has been produced.

https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#HookProduceRecordUnbuffered

https://github.com/twmb/franz-go/blob/8406043eb1a3eb5e117336b373106a5a9003c9f6/plugin/kotel/tracer.go#L184-L199

kafka/metrics.go
```diff
@@ -0,0 +1,103 @@
...
+	b, err := m.Int64Counter(
+		"write.error.count",
+		metric.WithDescription("The total number of error occurred on write"),
+		metric.WithUnit(unitCount),
+	)
+	if err != nil {
+		panic(err)
+	}
+
+	c, err := m.Int64Counter(
```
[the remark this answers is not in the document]
endorama: removed as suggested

kafka/metrics.go
endorama: Thanks for the hint, updated

kafka/metrics.go
```diff
@@ -0,0 +1,103 @@
...
+	"go.opentelemetry.io/otel/metric"
+)
+
+const instrumentName = "github.com/elastic/apm-queue/kafka"
+const unitCount = "1"
+
+type instruments struct {
+	MessageProduced metric.Int64Counter
+	WriteErrors     metric.Int64Counter
+	WriteTimeout    metric.Int64Counter
```
dmathieu, line 19: How about making this a `counters map[string]metric.Int64Counter`?
Then, you can store each generated metric within the map, and retrieve them by calling the proper key. And when we add more metrics, we only need to instantiate them, not to add new parameters to the struct.

Example: https://github.com/elastic/apm-ingest-service/pull/188/files#diff-8dcf09af3338b5e3276cbcefae7f6b12b8a4bde0662c242317cc543af8f95d15R63
And in otel's net/http: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/handler.go#L46

marclop: I'm not sure I agree. Seems unnecessary to add an unstructured map here since the metrics are well defined.

axw: I agree with @marclop, I don't see a good reason to use a map when we could use specific fields. The latter approach will naturally have lower overhead.

kafka/metrics.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+type kgoHooks struct {
+	instruments instruments
+}
+
+func NewKgoHooks(mp metric.MeterProvider) *kgoHooks {
```
marclop, line 43: Would you mind changing the returns for this function to be `(kgo.Hooks, error)`? I'd prefer to have a runtime error than having to call `panic(err)`  for every error.

kafka/metrics.go
```diff
@@ -0,0 +1,120 @@
...
+	)
+	if err != nil {
+		panic(err)
+	}
+
+	return &kgoHooks{
+		instruments{
+			MessageProduced: a,
+			WriteErrors:     b,
+			WriteTimeout:    c,
```
marclop, line 77: I find it hard to correlate the `a`, `b`, `c` variables with their instruments. Could you change the variable names to resemble the metric they represent?

marclop: Also, I'd prefer to a name `instrument` field, rather than specifying the `instruments` as the only struct argument.

kafka/metrics.go
```diff
@@ -0,0 +1,120 @@
...
+	"go.opentelemetry.io/otel/metric"
+)
+
+const instrumentName = "github.com/elastic/apm-queue/kafka"
+const unitCount = "1"
+
+type instruments struct {
+	MessageProduced metric.Int64Counter
+	WriteErrors     metric.Int64Counter
+	WriteTimeout    metric.Int64Counter
```
marclop, line 36: The struct is not exported, is there any reason to use exported fields in this struct? I think they can be un-exported

kafka/producer_metric_test.go
```diff
@@ -0,0 +1,143 @@
...
+	})
+	require.NoError(t, err)
+	require.NotNil(t, producer)
+
+	// NOTE: these tests are flaky
+	// Using go-cmp instead of metricdatatest.AssertEqual allow to bypass
+	// checkes on all metric data fields, which in some cases mey result in
+	// flakiness (i.e. attribute changing value).
+	// But that's not enough to remove all flakiness: as Producer is using 2
+	// meter provider, their order in the rm.ScopeMetrics slice may change,
```
marclop, line 75: We should probably filter out `ScopeMetrics.Scope == instrumentName` (`github.com/elastic/apm-queue/kafka`), then the test won't be flaky.

kafka/producer_metric_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+		assert.True(t, cmp.Equal(want, metric, opt))
+	})
+
+	t.Run("produced", func(t *testing.T) {
+		ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+		defer cancel()
+
+		err = producer.ProcessBatch(ctx, &model.Batch{
+			model.APMEvent{Transaction: &model.Transaction{ID: "1"}},
```
marclop, line 122: Can you change this test to produce multiple events?

kafka/metrics.go
```diff
@@ -0,0 +1,120 @@
...
+				context.Background(),
+				1,
+				metric.WithAttributeSet(attrs))
+		}
+		h.instruments.WriteErrors.Add(
+			context.Background(),
+			1,
+			metric.WithAttributeSet(attrs))
+		return
+	}
```
marclop, line 99: Rather than using two different metrics, do you think it makes sense to use dimensions for the different types of error. A timeout is still considered an error, it's just a type of error.

I'm thinking there could be three buckets:

- Timeout
- Canceled
- Other

This way we can aggregate all errors and / or filter where applicable.

kafka/metrics.go
```diff
@@ -0,0 +1,120 @@
...
+
+	h.instruments.MessageProduced.Add(
+		context.Background(),
+		1,
+		metric.WithAttributeSet(attrs))
+
+}
+
+// https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#HookBrokerWrite
+func (h *kgoHooks) OnBrokerWrite(meta kgo.BrokerMetadata, _ int16, _ int, writeWait, _ time.Duration, err error) {
```
marclop, line 109: Could you explain why we need this hook? Reading the description provided in https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#HookBrokerWrite, it seems that any write to a broker will call this hook. This includes pings, heartbeats, etc.

endorama: correct, removed

kafka/metrics.go
endorama: Nice catch, let me know if the new implementation works better.

kafka/metrics.go
endorama: no reason, unexported

kafka/metrics.go
endorama: I changed the names to more readable ones

kafka/metrics.go
endorama: added proper error handling

kafka/producer_metric_test.go
endorama: done

kafka/metrics.go
```diff
@@ -0,0 +1,105 @@
...
+type instruments struct {
+	messageProduced metric.Int64Counter
+	writeErrors     metric.Int64Counter
+}
+
+type kgoHooks struct {
+	instruments instruments
+}
+
+func NewKgoHooks(mp metric.MeterProvider) (*kgoHooks, error) {
```
marclop, line 48: It appears that this function isn't used outside of this package and I think the intended usage is internal.
```suggestion
func newKgoHooks(mp metric.MeterProvider) (*kgoHooks, error) {
```

kafka/metrics.go
```diff
@@ -0,0 +1,105 @@
...
+			metric.WithAttributes(partitionDimension, topicDimension, errorDimension),
+		)
+		return
+	}
+
+	h.instruments.messageProduced.Add(
+		context.Background(),
+		1,
+		metric.WithAttributes(partitionDimension, topicDimension))
+
```
marclop, line 104: WDYT?
```suggestion
	partition := attribute.Int("partition", int(r.Partition))
	topic := attribute.String("topic", r.Topic)

	if err != nil {
		errType := attribute.String("error", "other")
		if errors.Is(err, context.DeadlineExceeded) {
			errType = attribute.String("error", "timeout")
		}
		if errors.Is(err, context.Canceled) {
			errType = attribute.String("error", "cancelled")
		}

		h.instruments.writeErrors.Add(context.Background(), 1,
			metric.WithAttributes(partition, topic, errType),
		)
		return
	}
	h.instruments.messageProduced.Add(context.Background(), 1,
		metric.WithAttributes(partition, topic),
	)
```

kafka/common.go
```diff
@@ -178,9 +190,16 @@ func (cfg *CommonConfig) newClient(additionalOpts ...kgo.Opt) (*kgo.Client, erro
 	if !cfg.DisableTelemetry {
 		kotelService := kotel.NewKotel(
 			kotel.WithTracer(kotel.NewTracer(kotel.TracerProvider(cfg.tracerProvider()))),
-			kotel.WithMeter(kotel.NewMeter()),
+			kotel.WithMeter(kotel.NewMeter(kotel.MeterProvider(cfg.meterProvider()))),
+		)
+		metricHooks, err := NewKgoHooks(cfg.meterProvider())
+		if err != nil {
+			return nil, fmt.Errorf("cannot setup kgo metrics hooks: %w", err)
```
marclop, line 197: For consistency with other errors:
```suggestion
			return nil, fmt.Errorf("kafka: failed creating kgo metrics hooks: %w", err)
```

kafka/metrics.go
```diff
@@ -0,0 +1,105 @@
...
+	topicDimension := attribute.String("topic", r.Topic)
+
+	if err != nil {
+		errorDimension := attribute.String("error", "other")
+
+		if errors.Is(err, context.DeadlineExceeded) {
+			errorDimension = attribute.String("error", "timeout")
+		}
+		if errors.Is(err, context.Canceled) {
+			errorDimension = attribute.String("error", "cancelled")
```
marclop, line 89:
```suggestion
			errorDimension = attribute.String("error", "canceled")
```

kafka/metrics.go
```diff
@@ -0,0 +1,105 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+const (
+	instrumentName = "github.com/elastic/apm-queue/kafka"
+
+	unitCount = "1"
+
+	messageProducedCounterKey = "message.produced"
```
marclop, line 35: Should we prefix the metrics with the actual object that's producing them? For instance, this metric is generated by the producer, and `kafka.NewConsumer` won't affect these.

```suggestion
	producerMessagesProducedKey = "producer.messages.produced"
```

kafka/metrics.go
```diff
@@ -0,0 +1,105 @@
...
+	writeErrorCounterKey      = "write.error"
```
marclop, line 36:
```suggestion
	producerMessagesErrorKey      = "producer.messages.error"
```

kafka/producer_metric_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+func extractMetrics(t *testing.T, sm []metricdata.ScopeMetrics) []metricdata.Metrics {
+	t.Helper()
+
+	for _, m := range sm {
+		if m.Scope.Name == instrumentName {
+			return m.Metrics
+		}
+	}
+
```
marclop, line 109:
```suggestion
	t.Fatal("unable to find metrics for", instrumentName)
```

marclop (commented):
Looks pretty good! Left a few suggestions, mostly on naming and error conventions.

kafka/metrics.go
endorama: Agreed

endorama: Do we want this to be `producer` or `kafka`? Do you think it make sense to add a dimension to be able to filter for producer?

kafka/metrics.go
endorama: Added this manually in another commit, as could not apply 2 suggestions on the same code from the review screen.

kafka/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+	"go.opentelemetry.io/otel/metric"
+)
+
+const (
+	instrumentName = "github.com/elastic/apm-queue/kafka"
+
+	unitCount = "1"
+
+	messageProducedCounterKey = "producer.messages.produced"
+	writeErrorCounterKey      = "producer.messages.error"
```
marclop, line 36: What do you think about renaming this one to `errored`?

```suggestion
	writeErrorCounterKey      = "producer.messages.errored"
```

kafka/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+type instruments struct {
+	messageProduced metric.Int64Counter
+	writeErrors     metric.Int64Counter
```
marclop, line 41:
```suggestion
	messageErrored     metric.Int64Counter
```

kafka/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+	messageProducedCounter, err := m.Int64Counter(
+		messageProducedCounterKey,
+		metric.WithDescription("The total number of message produced"),
+		metric.WithUnit(unitCount),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create %s metric: %w", messageProducedCounterKey, err)
+	}
+
+	writeErrorCounter, err := m.Int64Counter(
```
marclop, line 60:
```suggestion
	messageErroredCounter, err := m.Int64Counter(
```

kafka/producer_metric_test.go
```diff
@@ -0,0 +1,149 @@
...
+func testMetric(t *testing.T, want metricdata.Metrics, actual metricdata.Metrics) {
+	t.Helper()
+
+	assert.Equal(t, want.Name, actual.Name)
+
+	wantValue := want.Data.(metricdata.Sum[int64]).DataPoints[0].Value
+	metricValue := actual.Data.(metricdata.Sum[int64]).DataPoints[0].Value
+	assert.Equal(t, wantValue, metricValue)
+}
```
marclop, line 121: Rather than using manual assertions, did you consider using [`go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest`](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest)? It seems more complete and it's easier to also assert dimensions. The current assertion completely dismisses dimensions.

kafka/producer_metric_test.go
```diff
@@ -0,0 +1,149 @@
...
+	"go.opentelemetry.io/otel/sdk/metric/metricdata"
+	"go.opentelemetry.io/otel/trace"
+	"go.uber.org/zap"
+
+	"github.com/elastic/apm-data/model"
+	apmqueue "github.com/elastic/apm-queue"
+	"github.com/elastic/apm-queue/codec/json"
+)
+
+func TestProducerMetrics_deadlineExceeded(t *testing.T) {
```
marclop, line 37: Ideally, I'd like to see 3 tests in this failure:

1. DeadlineExceeded
2. Canceled
3. Other error (we can close the broker before the batch is produced).

Then, we can assert that each of the metrics is grouped in the right dimension.

marclop (commented):
I think we should have a slightly more complete test and assert metric dimensions.

kafka/producer_metric_test.go
endorama: Given that the test was only targetting name and value using the entire metricdatatest was overkill and producing some issues, as there were fields conflicting then performing the equality check. I worked around that with `go-cmp` but in the end this was simpler and achieving the same goal.
I agree that if we want to test dimension this approach may be revised.

endorama: I added dimensions test, but now I introduced again flakiness due to the `partition` dimension. Is `partition` an useful dimension to use?

By reading `Record.Partition` [docs](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo?utm_source=gopls#Record.Partition) it reads:
> For producing, this is left unset. This will be set by the client before the record is unbuffered

So I'm thinking this may not be an interesting producer attribute. What do you think?

kafka/producer_metric_test.go
endorama: done

kafka/producer_metric_test.go
marclop: ah, what you're experiencing is caused by the fact that we have two partitions by default created by the helper `newClusterWithTopics`. I've opened a PR to change the signature of that function to require a partition number.

@axw I'm thinking it may be useful to see the producing throughput per partition rather than just on a per-topic aggregate. Do you think it's too much, or worth adding a dimension to it?

kafka/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+type kgoHooks struct {
+	instruments instruments
+}
+
+func newKgoHooks(mp metric.MeterProvider) (*kgoHooks, error) {
+	m := mp.Meter(instrumentName)
+
+	messageProducedCounter, err := m.Int64Counter(
+		messageProducedCounterKey,
+		metric.WithDescription("The total number of message produced"),
```
marclop, line 53:
```suggestion
		metric.WithDescription("The number of messages produced"),
```

kafka/producer_metric_test.go
```diff
@@ -0,0 +1,249 @@
...
+		Encoder: codec,
+		TopicRouter: func(event model.APMEvent) apmqueue.Topic {
+			return topic
+		},
+	})
+	require.NoError(t, err)
+	require.NotNil(t, producer)
+
+	return producer, rdr
+}
```
marclop, line 249: There's a lot of code duplication in these tests, I'm thinking we can simplify by creating a parent `TestProducerMetrics` and sub-tests for each of the dimensions and metrics we're asserting. I've made some changes locally after merging the current main and use a single partition for the test topic:

```suggestion
func TestProducerMetrics(t *testing.T) {
	test := func(ctx context.Context,
		t *testing.T,
		producer apmqueue.Producer,
		rdr sdkmetric.Reader,
		want metricdata.Metrics,
	) {
		err := producer.ProcessBatch(ctx, &model.Batch{
			model.APMEvent{Transaction: &model.Transaction{ID: "1"}},
			model.APMEvent{Transaction: &model.Transaction{ID: "2"}},
			model.APMEvent{Span: &model.Span{ID: "3"}},
		})
		assert.NoError(t, err)

		var rm metricdata.ResourceMetrics
		assert.NoError(t, rdr.Collect(context.Background(), &rm))

		metrics := filterMetrics(t, rm.ScopeMetrics)
		assert.Len(t, metrics, 1)
		metricdatatest.AssertEqual(t, want, metrics[0],
			metricdatatest.IgnoreTimestamp(),
		)
	}
	t.Run("DeadlineExceeded", func(t *testing.T) {
		producer, rdr := setupTestProducer(t)
		want := metricdata.Metrics{
			Name:        "producer.messages.errored",
			Description: "The number messages that failed to be produced",
			Unit:        "1",
			Data: metricdata.Sum[int64]{
				Temporality: metricdata.CumulativeTemporality,
				IsMonotonic: true,
				DataPoints: []metricdata.DataPoint[int64]{
					{
						Value: 3, Attributes: attribute.NewSet(
							attribute.String("error", "timeout"),
							attribute.String("topic", "default-topic"),
							attribute.Int("partition", 0),
						),
					},
				},
			},
		}
		ctx, cancel := context.WithTimeout(context.Background(), 0)
		defer cancel()
		test(ctx, t, producer, rdr, want)
	})
	t.Run("ContextCanceled", func(t *testing.T) {
		producer, rdr := setupTestProducer(t)
		want := metricdata.Metrics{
			Name:        "producer.messages.errored",
			Description: "The number messages that failed to be produced",
			Unit:        "1",
			Data: metricdata.Sum[int64]{
				Temporality: metricdata.CumulativeTemporality,
				IsMonotonic: true,
				DataPoints: []metricdata.DataPoint[int64]{
					{
						Value: 3, Attributes: attribute.NewSet(
							attribute.String("error", "canceled"),
							attribute.String("topic", "default-topic"),
							attribute.Int("partition", 0),
						),
					},
				},
			},
		}
		ctx, cancel := context.WithCancel(context.Background())
		cancel()
		test(ctx, t, producer, rdr, want)
	})
	t.Run("Other", func(t *testing.T) {
		producer, rdr := setupTestProducer(t)
		want := metricdata.Metrics{
			Name:        "producer.messages.errored",
			Description: "The number messages that failed to be produced",
			Unit:        "1",
			Data: metricdata.Sum[int64]{
				Temporality: metricdata.CumulativeTemporality,
				IsMonotonic: true,
				DataPoints: []metricdata.DataPoint[int64]{
					{
						Value: 3,
						Attributes: attribute.NewSet(
							attribute.String("error", "other"),
							attribute.String("topic", "default-topic"),
							attribute.Int("partition", 0),
						),
					},
				},
			},
		}
		require.NoError(t, producer.Close())
		test(context.Background(), t, producer, rdr, want)
	})
	t.Run("Produced", func(t *testing.T) {
		producer, rdr := setupTestProducer(t)
		want := metricdata.Metrics{
			Name:        "producer.messages.produced",
			Description: "The number of messages produced",
			Unit:        "1",
			Data: metricdata.Sum[int64]{
				Temporality: metricdata.CumulativeTemporality,
				IsMonotonic: true,
				DataPoints: []metricdata.DataPoint[int64]{
					{
						Value: 3,
						Attributes: attribute.NewSet(
							attribute.String("topic", "default-topic"),
							attribute.Int("partition", 0),
						),
					},
				},
			},
		}
		test(context.Background(), t, producer, rdr, want)
	})
}

func filterMetrics(t testing.TB, sm []metricdata.ScopeMetrics) []metricdata.Metrics {
	t.Helper()

	for _, m := range sm {
		if m.Scope.Name == instrumentName {
			return m.Metrics
		}
	}
	t.Fatal("unable to find metrics for", instrumentName)
	return []metricdata.Metrics{}
}

func setupTestProducer(t testing.TB) (*Producer, sdkmetric.Reader) {
	t.Helper()

	rdr := sdkmetric.NewManualReader()
	topic := apmqueue.Topic("default-topic")
	_, brokers := newClusterWithTopics(t, 1, topic)
	producer, err := NewProducer(ProducerConfig{
		CommonConfig: CommonConfig{
			Brokers:        brokers,
			Logger:         zap.NewNop(),
			TracerProvider: trace.NewNoopTracerProvider(),
			MeterProvider: sdkmetric.NewMeterProvider(
				sdkmetric.WithReader(rdr),
			),
		},
		Sync:    true,
		Encoder: json.JSON{},
		TopicRouter: func(event model.APMEvent) apmqueue.Topic {
			return topic
		},
	})
	require.NoError(t, err)
	require.NotNil(t, producer)
	return producer, rdr
}
```

kafka/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+		metric.WithUnit(unitCount),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create %s metric: %w", messageProducedCounterKey, err)
+	}
+
+	messageErroredCounter, err := m.Int64Counter(
+		messageErroredCounterKey,
+		metric.WithDescription("The total number of error occurred on write"),
```
marclop, line 62:
```suggestion
		metric.WithDescription("The number messages that failed to be produced"),
```

marclop (commented):
Just a couple more changes and it's ready to be merged!

kafka/metrics.go
```diff
@@ -0,0 +1,100 @@
...
+	messageErroredCounterKey  = "producer.messages.errored"
+)
+
+type instruments struct {
+	messageProduced metric.Int64Counter
+	messageErrored  metric.Int64Counter
+}
+
+type kgoHooks struct {
+	instruments instruments
```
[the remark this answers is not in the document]
endorama: Compared to adding them directly under `kgoHooks`? I would prefer to keep it like this as a sort of convention (grepping a codebase for `instruments` could be a nice way to quickly check where instrumentation is present).

But no hard feelings. What's your point of view?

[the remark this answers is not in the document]
endorama: that's true and I agree, thanks for the suggestion!

marclop (approved):
Looks great! Thanks for bearing with me on the reviews.

marclop:
@endorama it seems that the commits that you pushed aren't signed. Would you mind signing them?

dmathieu (approved):

PR #182, opened 2023-05-31
kafka: create separate TopicCreator type

axw:
This allows one to use a single Manager to create topics with differing partition counts and configs. It also enables one to create a Manager without specifying PartitionCount, e.g. in case one just wants to list topics.

### DIFF
--- a/kafka/topiccreator.go
+++ b/kafka/topiccreator.go
@@ -124,12 +124,15 @@ func (c *TopicCreator) CreateTopics(ctx context.Context, topics ...apmqueue.Topi
 		logger := c.m.cfg.Logger.With(zap.String("topic", response.Topic))
 		if err := response.Err; err != nil {
 			if errors.Is(err, kerr.TopicAlreadyExists) {
+				// NOTE(axw) apmotel currently does nothing with span events,
+				// hence we log as well as create a span event.
+				logger.Debug("kafka topic already exists")
 				span.AddEvent("kafka topic already exists", trace.WithAttributes(
 					semconv.MessagingDestinationKey.String(response.Topic),
 				))
 			} else {
 				span.RecordError(err)
-				span.SetStatus(codes.Error, "failed to create one or more topic")
+				span.SetStatus(codes.Error, err.Error())
 				createErrors = append(createErrors,
 					fmt.Errorf(
 						"failed to create topic %q: %w",
--- a/kafka/topiccreator.go
+++ b/kafka/topiccreator.go
@@ -48,9 +48,7 @@ type TopicCreatorConfig struct {
 	TopicConfigs map[string]string
 }
 
-// finalize ensures the configuration is valid, setting default values from
-// environment variables as described in doc comments, returning an error if
-// any configuration is invalid.
+// Validate ensures the configuration is valid, returning an error otherwise.
 func (cfg TopicCreatorConfig) Validate() error {
 	var errs []error
 	if cfg.PartitionCount == 0 {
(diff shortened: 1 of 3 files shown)


kafka/topiccreator.go
```diff
@@ -0,0 +1,143 @@
...
+	responses, err := c.m.client.CreateTopics(
+		ctx,
+		int32(c.partitionCount),
+		-1, // default.replication.factor
+		c.topicConfigs,
+		topicNames...,
+	)
+	if err != nil {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, "CreateTopics returned an error")
```
dmathieu, line 111: Why not report the actual error?

```suggestion
		span.SetStatus(codes.Error, err.Error())
```

kafka/topiccreator.go
```diff
@@ -0,0 +1,143 @@
...
+		createTopicParamsFields = append(createTopicParamsFields,
+			zap.Reflect("topic_configs", c.origTopicConfigs),
+		)
+	}
+	var createErrors []error
+	for _, response := range responses.Sorted() {
+		logger := c.m.cfg.Logger.With(zap.String("topic", response.Topic))
+		if err := response.Err; err != nil {
+			if errors.Is(err, kerr.TopicAlreadyExists) {
+				logger.Debug("kafka topic already exists")
```
dmathieu, line 127: How about adding an attribute to the span here, or an event?

kafka/topiccreator.go
axw: SGTM. I don't think we do anything with the status message at the moment, but perhaps we should.

kafka/topiccreator.go
```diff
@@ -0,0 +1,145 @@
...
+		createTopicParamsFields = append(createTopicParamsFields,
+			zap.Reflect("topic_configs", c.origTopicConfigs),
+		)
+	}
+	var createErrors []error
+	for _, response := range responses.Sorted() {
+		logger := c.m.cfg.Logger.With(zap.String("topic", response.Topic))
+		if err := response.Err; err != nil {
+			if errors.Is(err, kerr.TopicAlreadyExists) {
+				span.AddEvent("kafka topic already exists", trace.WithAttributes(
```
dmathieu, line 127: I just remembered: our agent has no notion of events.
So right now, apmotel doesn't do anything with them. While doing this seems like the right thing to do, it's currently going into `/dev/null`.

We should maybe do both logging and eventing?

kafka/topiccreator.go
```diff
@@ -0,0 +1,145 @@
...
+					semconv.MessagingDestinationKey.String(response.Topic),
+				))
+			} else {
+				span.RecordError(err)
+				span.SetStatus(codes.Error, "failed to create one or more topic")
```
dmathieu, line 132:
```suggestion
				span.SetStatus(codes.Error, err.Error())
```

kafka/topiccreator.go
axw: Ah, good point. In theory we could translate span events to logs in the agent, but for now I have added the log message back.

kafka/topiccreator.go
axw: Done.

kafka/topiccreator_test.go
```diff
@@ -0,0 +1,173 @@
...
+			"retention.ms": "123",
+		},
+	})
+	require.NoError(t, err)
+
+	var createTopicsRequest *kmsg.CreateTopicsRequest
+	cluster.ControlKey(kmsg.CreateTopics.Int16(), func(req kmsg.Request) (kmsg.Response, error, bool) {
+		createTopicsRequest = req.(*kmsg.CreateTopicsRequest)
+		return &kmsg.CreateTopicsResponse{
+			Version: 7,
```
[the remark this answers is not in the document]
axw: Thank you :)

PR #186, opened 2023-06-05
kafka: measure consumer group lag

axw:
Use kadm to measure consumer group lag, and report as OpenTelemetry metrics.

### DIFF
--- a/kafka/manager.go
+++ b/kafka/manager.go
@@ -24,7 +24,10 @@ import (
 
 	"github.com/twmb/franz-go/pkg/kadm"
 	"github.com/twmb/franz-go/pkg/kerr"
+	"github.com/twmb/franz-go/pkg/kgo"
+	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/metric"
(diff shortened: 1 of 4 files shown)


kafka/manager.go
```diff
@@ -64,18 +71,36 @@ func NewManager(cfg ManagerConfig) (*Manager, error) {
...
+	m := &Manager{
+		cfg:         cfg,
+		client:      client,
+		adminClient: kadm.NewClient(client),
+		tracer:      cfg.tracerProvider().Tracer("kafka"),
+	}
+
+	mp := cfg.meterProvider()
+	meter := mp.Meter("github.com/elastic/apm-queue/kafka")
+	consumerGroupLagMetric, err := meter.Int64ObservableGauge("kafka.consumer_group_lag")
```
lahsivjar, line 83: I think (not sure if that is better) if we are not going to have more than one observable metric, we can probably just use the `metric.WithInt64Callback` option in `meter.Int64ObservableGauge`.

lahsivjar (approved):

lahsivjar (approved):

kafka/manager.go
axw: I briefly looked at it, but I don't know how to unregister the callback when the manager is closed. I'm going to leave it for now.

PR #197, opened 2023-06-08
pubsublite(publisher): Record metrics, add testing

marclop:
This PR is the counterpart of #181 for the pubsublite package, with a bit of refactoring that was needed in order to be able test the metrics.

Introduces a new package `pubsubabs` which provides the necessary means to abstract the types in the `pubsub` package. Two new interfaces are introduced; `PublisherClient` implements the `pubsub.PublisherClient`, and `Publisher`, which returns a `PublishResult` interface that can be used to assert that our business logic works as expected in unit tests.

This allows us to start writing unit tests for the pubsublite.Publisher.

`telemetry.Producer` now implements the `pubsubabs.Publisher` interface and has been refactored from a function to a type.

`producer.messages.produced` and `produced.messages.errored` metrics are now recorded and reported using the configured OTel meter provider.

### DIFF
--- a/pubsublite/internal/pubsubabs/publisher.go
+++ b/pubsublite/internal/pubsubabs/publisher.go
@@ -1,0 +1,132 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// Package pubsubabs provides an abstraction layer over the `pubsub`
+// PublisherClient types to allow testing.
+package pubsubabs
+
+import (
+	"context"
+
+	"cloud.google.com/go/pubsub"
+)
+
+// ErrorStop defines the Stop and Error methods.
+type ErrorStop interface {
+	// Stop sends all remaining published messages and closes publish streams.
+	// Returns once all outstanding messages have been sent or have failed to be
+	// sent. Stop should be called when the client is no longer required.
+	Error() error
+	// Error returns the error that caused the publisher client to terminate. The
+	// error returned here may contain more context than PublishResult errors. The
+	// return value may be nil if Stop() was called.
+	Stop()
+}
+
+// PublisherClient represents a PubSub PublisherClient.
+type PublisherClient interface {
+	ErrorStop
+	Publish(ctx context.Context, msg *pubsub.Message) *pubsub.PublishResult
+}
+
+// Publisher defines an abstracted interface that can be used for mocking or
+// testing purposes. It wraps a Publisher client.
+type Publisher interface {
+	ErrorStop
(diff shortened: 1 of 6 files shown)


pubsublite/producer.go
```diff
@@ -177,17 +184,16 @@ func (p *Producer) ProcessBatch(ctx context.Context, batch *model.Batch) error {
 		if err != nil {
 			return fmt.Errorf("pubsublite: failed to get publisher: %w", err)
 		}
+		if !p.cfg.Sync {
+			ctx = queuecontext.DetachedContext(ctx)
+		}
```
marclop, line 189: I fixed this along the way, otherwise spans may not be ended and / or metrics recorded. We do the same for Kafka:

https://github.com/elastic/apm-queue/blob/f95dfa46df6ff7867f26b33ddd79ee36be153ea9/kafka/producer.go#L218-L221

pubsublite/internal/pubsubabs/publisher.go
```diff
@@ -0,0 +1,132 @@
...
+	Get(ctx context.Context) (serverID string, err error)
+	// Get returns the server-generated message ID and/or error result of a Publish call.
+	// Get blocks until the Publish call completes or the context is done.
+	Ready() <-chan struct{}
+}
+
+// IPublishResult is a copy from the internal pubsub packgage. All the code
+// below has been copied from "cloud.google.com/go/internal/pubsub"
+// Copyright 2020 Google LLC
+type IPublishResult struct {
```
lahsivjar, line 94: It seems to be required only for a single test, what do you think about moving it to the test file?

marclop: Actually, it's not done in this PR but we'll need to use this type in the `pubsublite` package too, not only for that telemetry test. We could copy the type to each package `_test.go` files, or leave it here. Do you have any preferences?

lahsivjar: I see, I think it should be okay to leave it here then.

pubsublite/internal/telemetry/publisher.go
```diff
@@ -19,54 +19,132 @@ package telemetry
...
+type ProducerMetrics struct {
+	produced metric.Int64Counter
+	errored  metric.Int64Counter
+}
 
-	attrs = append(attrs,
-		semconv.MessagingSystemKey.String("pubsublite"),
-		semconv.MessagingDestinationKindTopic,
+// NewProducerMetrics instantiates the producer metrics.
+func NewProducerMetrics(mp metric.MeterProvider) (ProducerMetrics, error) {
```
lahsivjar, line 50: Suggestion: We can make this an optional config, for prod, we can use the global meter provider (set as default by the config) and for tests we can pass in our own.

pubsublite/internal/pubsubabs/publisher.go
marclop: I'll resolve this, then.

pubsublite/internal/telemetry/publisher.go
```diff
@@ -19,54 +19,132 @@ package telemetry
...
+	producedKey = "producer.messages.produced"
+	erroredKey  = "producer.messages.errored"
+
+	instrumentName = "github.com/elastic/apm-queue/pubsublite"
+)
 
-// Publisher adds telemetry data to messages published
-func Publisher(ctx context.Context, tracer trace.Tracer, msg *pubsub.Message, h producerHandler, attrs []attribute.KeyValue) *pubsub.PublishResult {
+// ProducerMetrics hold the metrics that are recorded for producers
+type ProducerMetrics struct {
```
lahsivjar, line 44: nit: the package name is `publisher`, going by that flow we can have it named `PublisherMetrics` (or vise versa)? (feel free to ignore)

lahsivjar (approved):
LGTM!

dmathieu (approved):

PR #200, opened 2023-06-09
kafka(producer) metrics: record.Headers attributes

marclop:
Updates the kafka producer metrics to include the record's headers as string attributes in metric dimensions. This is useful to include any metadata that may have been injected using the queue context in metrics.

### DIFF
--- a/kafka/metrics.go
+++ b/kafka/metrics.go
@@ -71,8 +71,14 @@ func newKgoHooks(mp metric.MeterProvider) (*metricHooks, error) {
 
 // https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#HookProduceRecordUnbuffered
 func (h *metricHooks) OnProduceRecordUnbuffered(r *kgo.Record, err error) {
-	partition := semconv.MessagingKafkaDestinationPartition(int(r.Partition))
-	topic := semconv.MessagingDestinationName(r.Topic)
+	attrs := make([]attribute.KeyValue, 0, 5) // Preallocate 5 elements.
+	attrs = append(attrs,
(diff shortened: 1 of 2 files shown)


kafka/metrics.go
```diff
@@ -71,8 +71,14 @@ func newKgoHooks(mp metric.MeterProvider) (*metricHooks, error) {
...
 func (h *metricHooks) OnProduceRecordUnbuffered(r *kgo.Record, err error) {
-	partition := semconv.MessagingKafkaDestinationPartition(int(r.Partition))
-	topic := semconv.MessagingDestinationName(r.Topic)
+	attrs := make([]attribute.KeyValue, 0, 5) // Preallocate 5 elements.
+	attrs = append(attrs,
+		semconv.MessagingDestinationName(r.Topic),
+		semconv.MessagingKafkaDestinationPartition(int(r.Partition)),
+	)
+	for _, v := range r.Headers {
+		attrs = append(attrs, attribute.String(v.Key, string(v.Value)))
```
dmathieu, line 80: Can't this be quite high cardinality, and generate a lot of data?

marclop: it shouldn't since headers are empty by default and should only be populated by whatever is specified by clients via `queuecontext.WithMetadata` call that sets a `map[string]string` in the headers.

https://github.com/elastic/apm-queue/blob/ab0f35f82688590a2d59e9411c003dea0c8bc323/kafka/producer.go#L193-L201

 It could be high cardinality if a huge map with request based data that changes on a per-request basis. Something I just realized is that I didn't exclude the `traceparent` header from being added as a dimension. I fixed it in 129f0c9

dmathieu (approved):

PR #207, opened 2023-06-19
Introduce `Record` and `Processor`, remove `Codec`

marclop:
This change transforms this module into a generic consume/produce to Kafka or PubSub Lite.
It decouples record encoding from producing, decoding from consuming and removes the inter-module dependency on `apm-data`.

Introduces a new top level `apmqueue.Record`, wrapping a slice of bytes (`[]byte`) with an `apmqueue.Topic`. The producer also interface is also changed to:

    Produce(context.Context, ...Record) error

A new top level `apmqueue.Producer` is also introduced which is accepted by the two consumer implementations.

    Process(context.Context, ...Record) error

Last it removes the `cmd.loadgen` since it's not as useful since any data can be produced to the queue, not only `model.Batch`.

`codec/json` is also removed since it's been moved to `apm-data`.


kafka/consumer.go
```diff
@@ -95,15 +86,14 @@ func (cfg *ConsumerConfig) finalize() error {
...
-	if cfg.Decoder == nil {
-		errs = append(errs, errors.New("kafka: decoder must be set"))
-	}
 	if cfg.Processor == nil {
 		errs = append(errs, errors.New("kafka: processor must be set"))
 	}
 	return errors.Join(errs...)
 }
 
+var _ apmqueue.Consumer = &Consumer{}
```
lahsivjar, line 95: nit (feel free to ignore): I assume this it to check if `apmqueue.Consumer` is implemented by *Consumer. Maybe we can do something like below:

```suggestion
var _ apmqueue.Consumer = (*Consumer)(nil)
```

kafka/consumer.go
```diff
@@ -78,7 +69,7 @@ type ConsumerConfig struct {
 	//
 	// The processing time of each processing cycle can be calculated as:
 	// record.process.time * MaxPollRecords.
-	Processor model.BatchProcessor
+	Processor apmqueue.Processor
```
lahsivjar, line 72: The comment needs an update.

queue.go
```diff
@@ -57,50 +55,47 @@ type Consumer interface {
...
-
-// TopicRouter is used to determine the destination topic for an model.APMEvent.
-type TopicRouter func(event model.APMEvent) Topic
+// Record wraps a record's value with the topic where it's produced / consumed.
+type Record struct {
+	// Topics holds the topic where the record will be produced.
+	Topic Topic
+	// Value holds the record's content. It must not be mutated after Produce.
+	Value []byte
+}
```
lahsivjar, line 82: (Followup): Do you think we should add a field for timestamp to record the time at which the event was published to the queue?

kafka/producer.go
```diff
@@ -130,7 +110,9 @@ func (cfg *ProducerConfig) finalize() error {
 	return errors.Join(errs...)
 }
 
-// Producer is a model.BatchProcessor that publishes events to Kafka.
+var _ apmqueue.Producer = &Producer{}
```
lahsivjar, line 113: nit (feel free to ignore): I assume this it to check if `apmqueue.Producer` is implemented by `*Producer`. Maybe we can do something like below:

```suggestion
var _ apmqueue.Producer = (*Producer)(nil)
```

kafka/producer_test.go
```diff
@@ -253,34 +220,17 @@ func TestProducerGracefulShutdown(t *testing.T) {
 		defer cancel()
 
 		// This is a workaround to hook into the processing code using json marshalling.
-		// The goal is try to close the producer before the batch is sent to kafka but after the ProcessBatch
+		// The goal is try to close the producer before the batch is sent to kafka but after the Produce
 		// method is called.
-		// We are using two goroutines because both ProcessBatch and Close will block waiting on each other.
+		// We are using two goroutines because both Produce and Close will block waiting on each other.
```
lahsivjar, line 225:
```suggestion
```
IIUC, the above line is no longer valid.

lahsivjar (commented):
Mostly LGTM, a few minor comments.

kafka/consumer.go
marclop: Why does it need to be updated? The function hasn't changed.

queue.go
marclop: What's the purpose of the timestamp field and how would this timestamp field be used? Is the goal to use it in `Config.Processor`?

kafka/consumer.go
lahsivjar:
```
	// Processor that will be used to process each record individually.
	// It is recommended to keep the synchronous processing fast and below the
	// rebalance.timeout.ms setting in Kafka.
```
Maybe(?)
Also, not part of the PR, I don't quite get the `rebalance.timeout.ms` point, maybe there is a typo there.

marclop: The processor will process each of the fetched record(s) individually. The comment is meant to inform users that passing a Processor which takes longer to process a single record than the rebalancing timeout in Kafka will potentially cause the record to be consumed twice (if a rebalance is waiting to happen while a record is being processed).

queue.go
lahsivjar: > What's the purpose of the timestamp field and how would this timestamp field be used?

One use case can be to get the queued delay (the time for which a record was in a queued state before being consumed).

marclop: Would you mind opening an issue for this enhancement? It seems like something we may want to implement, but I'm reluctant to add a field there until we have a legitimate reason to.

kafka/consumer.go
lahsivjar: Ah, thanks for the explanation. Now I can read the same line correctly, for some reason I kept reading `below` to mean beneath the line😅

queue.go
lahsivjar: 👍

queue.go
```diff
@@ -57,50 +55,47 @@ type Consumer interface {
...
 	Close() error
 }
 
-// Topic represents a destination topic where to produce a message/record.
-type Topic string
-
-// TopicRouter is used to determine the destination topic for an model.APMEvent.
-type TopicRouter func(event model.APMEvent) Topic
+// Record wraps a record's value with the topic where it's produced / consumed.
+type Record struct {
```
marclop, line 77: Instead of enriching the context with a map via `queuecontext.WithMetadata(ctx, map[string]string{}`, each record could have its own header, then it's up to the client to use the same headers for all the messages or have individual headers for each message.

kafka/producer_test.go
```diff
@@ -207,32 +188,18 @@ func TestNewProducerBasic(t *testing.T) {
...
-	return []byte("null"), nil
-}
-
 func TestProducerGracefulShutdown(t *testing.T) {
 	test := func(t testing.TB, dt apmqueue.DeliveryType, syncProducer bool) {
 		_, brokers := newClusterWithTopics(t, 1, "topic")
-		var codec json.JSON
 		var processed atomic.Int64
+		wait := make(chan struct{})
+		// signal := make(chan struct{})
```
axw, line 196:
```suggestion
```

axw (approved):

kyungeunni (approved):
LGTM!

PR #208, opened 2023-06-19
ci: correctly pass github token to workflow

amannocci:
## What is the change being made?

* Properly pass github token to workflow

## Why is the change being made?

* The GitHub token wasn't validated until last week.

### DIFF
--- a/.github/workflows/system-test.yml
+++ b/.github/workflows/system-test.yml
@@ -19,6 +19,8 @@ jobs:
       pull-requests: read
     steps:
       - uses: elastic/apm-pipeline-library/.github/actions/is-pr-approved@current
+        with:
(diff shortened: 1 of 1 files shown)


axw (approved):

.github/workflows/system-test.yml
```diff
@@ -19,6 +19,8 @@ jobs:
       pull-requests: read
     steps:
       - uses: elastic/apm-pipeline-library/.github/actions/is-pr-approved@current
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
reakaleek, line 23: Is this something like a predefined secret? I cannot find it in https://github.com/elastic/apm-queue/settings/secrets/actions

reakaleek (approved):

PR #212, opened 2023-06-20
pubsublite/consumer: Record `messages.processed`

### LINKED ISSUE #211: pubsublite: Produce the same consumer metrics as the kafka package
## Description

The `pubsublite` package does not publish any metrics for consuming. The kafka and pubsublite implementations should produce the same metrics with the same attributes.

marclop:
Records a new `consumer.messages.processed` metric in the pubsublite consumer.

Closes #211


pubsublite/internal/telemetry/consumer.go
```diff
@@ -21,42 +21,86 @@ package telemetry
...
+		return cm, fmt.Errorf(
+			"telemetry: cannot create 'consumer.messages.fetched' metric: %w", err,
+		)
+	}
+	return
+}
 
-// Consumer adds telemetry data to messages received
-func Consumer(tracer trace.Tracer, h consumerHandler, attrs []attribute.KeyValue) consumerHandler {
+// Consumer decorates an existing publisher with tracing and metering.
```
axw, line 57:
```suggestion
// Consumer decorates an existing consumer with tracing and metering.
```

pubsublite/internal/telemetry/consumer.go
```diff
@@ -21,42 +21,86 @@ package telemetry
...
-
-		attrs = append(attrs,
-			semconv.MessagingSystemKey.String("pubsublite"),
-			semconv.MessagingSourceKindTopic,
-			semconv.MessagingOperationProcess,
-			semconv.MessagingMessageIDKey.String(msg.ID),
-		)
-
-		ctx, span := tracer.Start(ctx, "pubsublite.Receive",
+		ctx, span := tracer.Start(ctx, "pubsublite.Receive", // PubSub name.
```
axw, line 93: Should we follow the OTel semantic conventions for the span name (https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/messaging/#span-name), and set it to `<topic> process?`

pubsublite/internal/telemetry/consumer.go
```diff
@@ -21,42 +21,86 @@ package telemetry
...
 			trace.WithSpanKind(trace.SpanKindConsumer),
-			trace.WithAttributes(attrs...),
+			trace.WithAttributes(append(attrs,
+				semconv.MessagingMessageIDKey.String(msg.ID),
```
axw, line 96: Should we also include `messaging.message.payload_size_bytes`, set to the length of msg.Data?

pubsublite/internal/telemetry/consumer.go
```diff
@@ -21,42 +21,86 @@ package telemetry
...
-			semconv.MessagingSourceKindTopic,
-			semconv.MessagingOperationProcess,
-			semconv.MessagingMessageIDKey.String(msg.ID),
-		)
-
-		ctx, span := tracer.Start(ctx, "pubsublite.Receive",
+		ctx, span := tracer.Start(ctx, "pubsublite.Receive", // PubSub name.
 			trace.WithSpanKind(trace.SpanKindConsumer),
-			trace.WithAttributes(attrs...),
+			trace.WithAttributes(append(attrs,
```
axw, line 95: `attrs` could in theory be the original `commonAttrs`, could this mutate the original slice?

axw (commented):
Looks good, just a few small things. I'm a bit concerned about the amount of overhead we're introducing by tracing every single message, but we can iterate on this.

pubsublite/internal/telemetry/consumer.go
marclop: I suppose it could if there's more capacity than length. I'll have a look at this.

kyungeunni (approved):

pubsublite/internal/telemetry/consumer.go
marclop: It seems we're always passing down a slice with `cap == len`, so this will never happen in practice. Would you prefer I copy the slice for each of the events? For now, I've added a test to cover this scenario.

axw: How about w reslice `commonAttrs` on line 77 to ensure `cap == len`, rather than relying on the caller doing that. i.e. `attrs := commonAttrs[0:len(commonAttrs):len(commonAttrs)]`

marclop: sounds good! I should've thought of that.

axw (approved):

PR #214, opened 2023-06-21
systemtest: remove nested go module

axw:
The dependencies of systemtest are contained by apm-queue, so remove the nested module. This will allow Renovate to do its thing.


marclop:
Won't this cause the `systemtest` to be run with `go test ./...` which are meant to run unit tests?

axw:
@marclop oof, yeah :/

That seems like kind of an indirect way to stop `go test ./...` from running unnecessary tests though. I'll have have a think about how we can improve this.

marclop:
the easiest way would be to modify the `test` Make target and use something like:

```
go test -race -v -timeout=$(GO_TEST_TIMEOUT) $(shell go list ./... | grep -v systemtest)
```

axw:
@marclop I'm thinking of making the system tests opt-in, so we can also run `go test ./...` on our laptops easily. That would require running `go test -run-system-tests` when in the system test directory. Slightly more painful to run system tests on laptops, but I think not too bad. WDYT?

marclop:
That seems a bit more friendly so we can keep using the `go test` commands directly without relying on the makefile necessarily.

marclop (approved):

PR #216, opened 2023-06-21
Add event delay metric on both producers

dmathieu:
This adds a new header on each producer record, `timestamp`, which stores the time when the record was produced.
It also emits a new metric, `consumer.messages.delay`, which is the delay in milliseconds it took to consume a record.

### DIFF
--- a/kafka/metrics.go
+++ b/kafka/metrics.go
@@ -137,6 +137,9 @@ func (h *metricHooks) OnFetchRecordUnbuffered(r *kgo.Record, _ bool) {
 			}
 
 			delay := time.Since(since).Seconds()
+			span.SetAttributes(
+				attribute.Float64(apmqueue.EventTimeKey, delay),
+			)
 			h.messageDelay.Record(r.Context, delay, metric.WithAttributes(
 				attrs...,
 			))
--- a/kafka/metrics.go
+++ b/kafka/metrics.go
@@ -21,11 +21,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/twmb/franz-go/pkg/kgo"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
 	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
+
+	apmqueue "github.com/elastic/apm-queue"
 )
 
 const (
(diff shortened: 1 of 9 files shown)


pubsublite/internal/telemetry/consumer.go
```diff
@@ -94,7 +112,17 @@ func Consumer(
 		metrics.fetched.Add(context.Background(), 1, metric.WithAttributes(
 			attrs...,
 		))
+
 		if msg.Attributes != nil {
+			if msg.Attributes[apmqueue.EventTimeKey] != "" {
+				if since, err := time.Parse(time.RFC3339, msg.Attributes[apmqueue.EventTimeKey]); err == nil {
+					delay := time.Since(since).Milliseconds()
```
lahsivjar, line 119: Unit issue, same as what was raised in previous comment.

kafka/metrics.go
```diff
@@ -67,12 +72,22 @@ func newKgoHooks(mp metric.MeterProvider) (*metricHooks, error) {
 	if err != nil {
 		return nil, formatMetricError(msgFetchedKey, err)
 	}
+
+	messageDelayHistogram, err := m.Int64Histogram(msgDelayKey,
+		metric.WithDescription("The delay between producing messages and reading them"),
+		metric.WithUnit(unitCount),
```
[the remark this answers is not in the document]
dmathieu: You're right, I missed it.

kafka/metrics.go
```diff
@@ -105,9 +120,22 @@ func (h *metricHooks) OnProduceRecordUnbuffered(r *kgo.Record, err error) {
...
+	attrs := attributesFromRecord(r)
+
 	h.messageFetched.Add(context.Background(), 1,
-		metric.WithAttributes(attributesFromRecord(r)...),
+		metric.WithAttributes(attrs...),
 	)
+
+	for _, v := range r.Headers {
+		if v.Key == apmqueue.EventTimeKey {
+			if since, err := time.Parse(time.RFC3339, string(v.Value)); err == nil {
```
lahsivjar, line 131: Should we log the error, it might help us down the line if something bad happens.

lahsivjar (approved):
LGTM! A minor comment on logging error.

kafka/metrics.go
dmathieu: Good call.
We don't have a logger within any of the telemetry packages. So I've recorded a span error.

carsonip (approved):

marclop (commented):
Is there a reason we're using an additional header to propagate when the record was produced to the queue? Both Kafka and PubSubLite have built-in fields for that purpose. We could be using `record.Timestamp` for Kafka, and `message.PublishTime` for PubSub Lite.

dmathieu:
You're right. I missed that while digging through the docs :(
https://github.com/elastic/apm-queue/pull/222

PR #220, opened 2023-06-22
Add MonitorConsumerLag for both kafka and pubsublite

### LINKED ISSUE #202: pubsublite: measure subscription lag
Similar to https://github.com/elastic/apm-queue/pull/186, we should measure Pub/Sub Lite subscription lag.

carsonip:
- Add monitoring for backlog_message_count in pubsublite like kafka.consumer_group_lag
- Only monitor consumer lag when explicitly called, for both kafka and pubsublite

Closes #202


pubsublite/manager.go
```diff
@@ -63,12 +74,33 @@ func NewManager(cfg ManagerConfig) (*Manager, error) {
...
+	m := &Manager{cfg: cfg, client: client}
+
+	m.monitoringClient, err = monitoring.NewMetricClient(context.Background())
+	if err != nil {
+		return nil, fmt.Errorf("pubsublite: failed creating monitoring client: %w", err)
+	}
+
+	mp := cfg.meterProvider()
+	meter := mp.Meter("github.com/elastic/apm-queue/pubsublite")
+	consumerGroupLagMetric, err := meter.Int64ObservableGauge("pubsublite.backlog_message_count")
```
carsonip, line 86: [for reviewers] I know that @axw is keen on having the same key for lag across kafka and pubsublite instead of having "kafka.consumer_group_lag" and "pubsublite.backlog_message_count". The labels will be different though, pubsublite has subscription_id while kafka has group + topic. Will that be a problem?

pubsublite/manager.go
```diff
@@ -240,3 +272,56 @@ func (m *Manager) DeleteSubscription(ctx context.Context, subscription string) e
 	logger.Info("deleted pubsublite subscription")
 	return nil
 }
+
+func (m *Manager) gatherMetrics(ctx context.Context, o metric.Observer) error {
+	it := m.monitoringClient.ListTimeSeries(ctx, &monitoringpb.ListTimeSeriesRequest{
+		Name:   fmt.Sprintf("projects/%s", m.cfg.Project),
+		Filter: "metric.type = \"pubsublite.googleapis.com/subscription/backlog_message_count\"",
+		Interval: &monitoringpb.TimeInterval{
+			StartTime: &timestamppb.Timestamp{Seconds: time.Now().Add(-5 * time.Minute).Unix()},
```
carsonip, line 281: [for reviewer] I tried starttime == endtime and it didn't like it. Not even with start time = now -1m, so I just set start time = now -5m and it just works.

pubsublite/manager.go
axw: Yes that would defeat the purpose. We encode Pub/Sub Lite subscription names with abstract notion of "consumer group": https://github.com/elastic/apm-queue/blob/fa8e0138acdfb9937e61e956ed83eb1da286d14e/pubsublite/common.go#L135-L139

We should reverse that here, and treat the bit after the "+" as the consumer group name.

pubsublite/manager.go
```diff
@@ -240,3 +272,56 @@ func (m *Manager) DeleteSubscription(ctx context.Context, subscription string) e
 	logger.Info("deleted pubsublite subscription")
 	return nil
 }
+
+func (m *Manager) gatherMetrics(ctx context.Context, o metric.Observer) error {
+	it := m.monitoringClient.ListTimeSeries(ctx, &monitoringpb.ListTimeSeriesRequest{
+		Name:   fmt.Sprintf("projects/%s", m.cfg.Project),
+		Filter: "metric.type = \"pubsublite.googleapis.com/subscription/backlog_message_count\"",
```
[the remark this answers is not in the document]
axw: Yeah that doesn't seem ideal. Can we filter by region? https://cloud.google.com/monitoring/api/resources#tag_pubsublite_subscription_partition says that the "location" is available as a label.

We could then cross reference the returned subscription metrics with the result of `ListReservationTopics`.

I think what we should do instead though is enable the caller to pass in a list of topics. Rather than registering the metric in NewManager, we would instead provide a method which accepts a list of topics and registers a callback. Something like this:

```go
// MonitorConsumerLag registers a callback with OpenTelemetry
// to measure consumer group lag for the given topics.
func (m *Manager) MonitorConsumerLag(topics ...apmqueue.Topic) (metric.Registration, error)
```

Then we could either include those topic names as prefix filters on the subscription IDs, or filter client side.

pubsublite/manager.go
carsonip: Unified the metric name. What about the meter name?
```
meter := mp.Meter("github.com/elastic/apm-queue/pubsublite")
```

pubsublite/manager.go
```diff
@@ -240,3 +252,93 @@ func (m *Manager) DeleteSubscription(ctx context.Context, subscription string) e
 	logger.Info("deleted pubsublite subscription")
 	return nil
 }
+
+// MonitorConsumerLag registers a callback with OpenTelemetry
+// to measure consumer group lag for the given topics.
+func (m *Manager) MonitorConsumerLag(topics ...string) (metric.Registration, error) {
+	monitoringClient, err := monitoring.NewMetricClient(context.Background(), m.cfg.MonitoringClientOptions...)
```
carsonip, line 259: [for reviewer] monitoringClient is initialized in a closure and no Close is called when shutting down, since the lifecycle is controlled by caller. Practically I don't think it matters but open to suggestions.

pubsublite/manager.go
```diff
@@ -240,3 +252,93 @@ func (m *Manager) DeleteSubscription(ctx context.Context, subscription string) e
 	logger.Info("deleted pubsublite subscription")
 	return nil
 }
+
+// MonitorConsumerLag registers a callback with OpenTelemetry
+// to measure consumer group lag for the given topics.
+func (m *Manager) MonitorConsumerLag(topics ...string) (metric.Registration, error) {
```
carsonip, line 258: It takes `string` for topic for consistency as other public methods under pubsublite manager also take a string topic, as opposed to apmqueue.Topic in Kafka manager.

axw: Actually we should probably accept "topics-consumers", i.e. a list of (topic, consumer) pairs. Topics may be prefixed for namespacing/multi-tenancy, whereas consumers will probably not differ across tenants.

kafka/manager.go
```diff
@@ -146,83 +127,103 @@ func (m *Manager) DeleteTopics(ctx context.Context, topics ...apmqueue.Topic) er
...
-	if err != nil {
-		span.RecordError(err)
-		span.SetStatus(codes.Error, err.Error())
-		return fmt.Errorf("failed to describe groups: %w", err)
+// MonitorConsumerLag registers a callback with OpenTelemetry
+// to measure consumer group lag for the given topics.
+func (m *Manager) MonitorConsumerLag(topics ...apmqueue.Topic) (metric.Registration, error) {
+	monitorTopics := make(map[string]struct{}, len(topics))
+	for _, t := range topics {
+		monitorTopics[string(t)] = struct{}{}
```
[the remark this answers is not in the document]
axw: We might want to pass in both, see comments on pubsublite.

kafka/manager.go
```diff
@@ -146,83 +127,106 @@ func (m *Manager) DeleteTopics(ctx context.Context, topics ...apmqueue.Topic) er
...
-	listPartitions := groups.AssignedPartitions()
-	listPartitions.Merge(commits.CommittedPartitions())
-	if topics := listPartitions.Topics(); len(topics) > 0 {
-		res, err := m.adminClient.ListEndOffsets(ctx, topics...)
+
+	gatherMetrics := func(ctx context.Context, o metric.Observer) error {
+		ctx, span := m.tracer.Start(ctx, "GatherMetrics")
+		defer span.End()
+
+		// Fetch commits for consumer groups.
```
axw, line 157: Can we pass in the consumer group names here? That will avoid listing all groups under the hood.

axw (commented):
Thanks, looks good. Just a couple of minor things.

I think TopicConsumer might end up getting removed when we address https://github.com/elastic/apm-queue/issues/224. Instead of passing in TopicConsumers, we would add a `Namespace string` field to `pubsublite.CommonConfig` and `kafka.CommonConfig`, and they would use that for prefixing (and filtering) topic names.

pubsublite/manager.go
```diff
@@ -22,16 +22,30 @@ import (
...
+
+	apmqueue "github.com/elastic/apm-queue"
 )
 
 // ManagerConfig holds configuration for managing GCP Pub/Sub Lite resources.
 type ManagerConfig struct {
 	CommonConfig
+
+	// MonitoringClientOptions holds arbitrary Google monitoring API client options.
+	MonitoringClientOptions []option.ClientOption
```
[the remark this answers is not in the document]
carsonip: It is for connecting to the mock monitoring api server, which is currently running separate to pubsub api server

kafka/manager.go
carsonip: Done!

pubsublite/manager.go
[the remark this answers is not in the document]
carsonip: Done!

pubsublite/manager.go
axw: We don't do anything with the meter name, so that's fine as it is for now.

kafka/manager_test.go
```diff
@@ -337,11 +356,9 @@ func TestManagerMetrics(t *testing.T) {
 		}},
 	}, metrics[0].Data, metricdatatest.IgnoreTimestamp())
 
-	assert.Equal(t, &kmsg.ListGroupsRequest{Version: 4}, listGroupsRequest)
-	assert.Equal(t, &kmsg.DescribeGroupsRequest{
-		Version: 5,
-		Groups:  []string{"connect", "consumer1", "consumer2", "consumer3"},
-	}, describeGroupsRequest)
+	assert.Nil(t, listGroupsRequest)
```
axw, line 359: Should we remove the `cluster.ControlKey(kmsg.ListGroups.Int16()...`? I don't think we need to test which APIs we're not calling.

axw (approved):
Thank you!

PR #223, opened 2023-06-24
Introduce ordering key for kafka and pubsublite

lahsivjar:
The ordering key allows Kafka and Pubsublite to ensure that a record with the same ordering key is routed to the same partition.

Franz-go library, by default, uses a [UniformBytesPartitioner](https://github.com/twmb/franz-go/blob/6147cee6468484e5dfab6febf4b76962dffaee38/pkg/kgo/config.go#L499) with ordering keys enabled and default murmur2 hasher. The `UniformBytesPartitioner` with ordering keys enabled uses the configured hash function to map a record with `Key` to a specific partition ([ref](https://github.com/twmb/franz-go/blob/6147cee6468484e5dfab6febf4b76962dffaee38/pkg/kgo/partitioner.go#L323-L325)). This means that we can keep using the default partitioner and satisfy the requirement for ordering keys.

### DIFF
--- a/kafka/producer.go
+++ b/kafka/producer.go
@@ -204,6 +204,7 @@ func (p *Producer) Produce(ctx context.Context, rs ...apmqueue.Record) error {
 		kgoRecord := &kgo.Record{
 			Headers: headers,
 			Topic:   string(record.Topic),
+			Key:     record.OrderingKey,
 			Value:   record.Value,
 		}
 		p.client.Produce(ctx, kgoRecord, func(r *kgo.Record, err error) {
--- a/systemtest/produce_consume_basic_test.go
+++ b/systemtest/produce_consume_basic_test.go
@@ -80,7 +80,12 @@ func TestProduceConsumeMultipleTopics(t *testing.T) {
 func TestProduceConsumeOrderingKeys(t *testing.T) {
 	// The test asserts that messages are consumed in sequence given that
 	// they are produced with a specific ordering key.
-	events := 50
+
+	// Use a high event count to circumvent the UniformBytesPartitioner
+	// used by franz-go by default which aims to produce large batches of
+	// data to same partition.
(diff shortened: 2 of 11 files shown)


kafka/producer.go
```diff
@@ -204,6 +204,7 @@ func (p *Producer) Produce(ctx context.Context, rs ...apmqueue.Record) error {
 		kgoRecord := &kgo.Record{
 			Headers: headers,
 			Topic:   string(record.Topic),
+			Key:     record.OrderingKey,
```
kyungeunni, line 207: It's not a comment directly related to this change, but why are we having this option? AFAIK we haven't used the key to route messages to a partition to avoid hot-spotting. I'm curious to know what use cases we'd need this(stateful processing?)

kyungeunni (approved):

kafka/producer.go
lahsivjar: We have a use case where we want a set of messages to be consumed by the same instance of the consumer.

systemtest/produce_consume_basic_test.go
```diff
@@ -126,6 +193,33 @@ func assertProcessor(t testing.TB, assertions consumerAssertions) apmqueue.Proce
 	})
 }
 
+// sequenceConsumerAssertions asserts that records are consumed in sequence
+// assuming that all records are sent to same partitions. If records are
+// not sent to same partition then it will result in failures.
+type sequenceConsumerAssertions struct {
+	consumerAssertions
```
marclop, line 200: seems we're only using the `records` field, should it just use `records *atomic.Int64`?

marclop (approved):
LGTM

systemtest/produce_consume_basic_test.go
lahsivjar: Done

PR #227, opened 2023-06-28
Update notice template

carsonip:


carsonip:
updated it by hand. Not sure why `make notice` doesn't work on my laptop today

PR #233, opened 2023-07-11
Don't run system-tests on PRs from forks

reakaleek:
## Details

- Removes the ability to run system tests on a PR from a forked repository.
- A successful status check is created without running the tests if it's a PR from a forked repository.
- A PR comment is created that warns about this behavior


.github/workflows/system-test-notification.yml
```diff
@@ -0,0 +1,32 @@
...
+    permissions:
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add comment
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const { repo, owner } = context.repo
+            await github.rest.issues.createComment({
```
reakaleek, line 23: @axw 

Unfortunately, it was not able to use a status check, because the token does not have access to create a commit status on a forked PR.

This is why I just create a comment instead.

axw: No problem, thanks for explaining

axw (approved):

.github/workflows/system-test-notification.yml
```diff
@@ -0,0 +1,32 @@
+name: system-test-notification
+
+on:
+  # This needs to be `pull_request_target` instead of `pull_request`
+  # to be able to write a comment on the pull request.
+  pull_request_target:
+    types:
+      - opened
```
kuisathaverat, line 8: a ` pull_request:` workflow can write comments in PRs, what gives it permissions is the `permisions` setting

.github/workflows/system-test-notification.yml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  notify:
+    # This job runs only if the pull request is from a forked repository.
+    if: github.event.pull_request.head.repo.full_name != github.repository
```
kuisathaverat, line 13: why not use `github.event.pull_request.head.repo.fork == true`?

.github/workflows/system-test-notification.yml
reakaleek: Let me verify this again. But yesterday, when I tried it - with the `permissions` already set, I got  a `403` error.

reakaleek: See this [PR](https://github.com/unelastisch/actions-playground/pull/20) in my personal repository.

The notification is failing with a `403` error in this [workflow run](https://github.com/unelastisch/actions-playground/actions/runs/5541288320/jobs/10114495237?pr=20).

.github/workflows/system-test-notification.yml
reakaleek: https://github.com/elastic/apm-queue/pull/233/commits/7cbd688bd21f851e771b043caeeb49b76d5b7776

.github/workflows/system-test-notification.yml
reakaleek: I think it's because the github token does not have any write permissions, even if you set it expliclity on a PR from a forked repo.

kuisathaverat: since the wf does not make any checkout from the PR, I think is fine

kuisathaverat (approved):

Issue #235, opened 2023-07-12
kafka: support loading Azure credentials from disk


axw:
A couple of ideas:
 1. we could revive https://github.com/elastic/apm-queue/pull/43
 2. we could add support for reading SAS credentials from a file, like in https://learn.microsoft.com/en-us/azure/event-hubs/azure-event-hubs-kafka-overview#shared-access-signature-sas

OAUTHBEARER (1) is probably best in the long run, but for now let's see what we can do around reloading SAS credentials (2). It doesn't need to be full-blown .properties JAAS format.

Issue #234, opened 2023-07-12
kafka: support loading AWS MSK credentials from ~/.aws


axw:
We should support loading credentials from disk, rather than environment variables. We could pull in https://github.com/aws/aws-sdk-go-v2, or we could add support for parsing the files to apm-queue. See https://docs.aws.amazon.com/sdkref/latest/guide/creds-config-files.html

Issue #236, opened 2023-07-12
Graceful reloading/reauthentication on credentials rotation


axw:
Ensure that both the kafka and pubsublite code gracefully handles credentials changes.

Related: https://github.com/elastic/apm-queue/issues/234 https://github.com/elastic/apm-queue/issues/235

axw:
At least in theory, the seed brokers list could change over time too. e.g. if migrating to a different Azure Event Hubs namespace. This can be achieved by calling kgo.Client.UpdateSeedBrokers.

axw:
For Kafka, I'm thinking of writing a kgo hook that will periodically load a file from disk, extracting select properties defined at https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md (namely `bootstrap.servers` and `sasl.{username,password}`), compare them to the last loaded values, and reconfigure the client if they have changed.

simitt:
@axw is the kafka part done with https://github.com/elastic/apm-queue/pull/241?

I think it would be best to re-focus this issue on `aws` and create a dedicated one for the work in `azure` EventHubs.

axw:
@simitt that PR handles both Kafka and Event Hubs. I don't think I tested Pub/Sub Lite reloading yet, so that would need a new issue.

simitt:
The part that is left to do here is updating each of the services to load the config from a file. @lahsivjar pointed out that https://github.com/knadh/koanf#reading-config-from-files might be a good choice.

axw:
Still need to verify GCP, will come back to that later.

PR #240, opened 2023-07-14
kafka: introduce config file option

axw:
Introduce CommonConfig.ConfigFile, which
should point to a YAML file with configuration.
The filename may also be specified with the
environment variable $KAFKA_CONFIG_FILE.

The config file can include:
 - bootstrap.servers
 - sasl.mechanism
 - sasl.username
 - sasl.password

If a config file is specified, a hook will be
added to kgo.Client to reload seed brokers from
the file when a broker connection fails. Thus,
changes to the file will be picked up when the
old broker addresses begin failing.

If `sasl.mechanism` is set to PLAIN, or
`sasl.username` is specified, then SASL/PLAIN
is configured and will reload the username and
password from the file on each authentication
attempt. Thus, changes to the credentials will
be picked up when connections are recreated.

Presence of the AWS_ACCESS_KEY_ID environment
variable no longer configures SASL/AWS_MSK_IAM
automatically. It is now necessary to specify
`KAFKA_SASL_MECHANISM=AWS_MSK_IAM`, or to specify
`sasl.mechanism: AWS_MSK_IAM` in the config file.
When this is set, we now use the AWS SDK to
confgure AWS_MSK_IAM; changes to credentials
files will be picked up on reauthentication.

Part of https://github.com/elastic/apm-queue/issues/236
Closes https://github.com/elastic/apm-queue/issues/234
Closes https://github.com/elastic/apm-queue/issues/235


axw:
Reopening on elastic repo.

Issue #224, opened 2023-06-26
Introduce `namespace` concept


marclop:
## Description

Some of the resources created by the topic / reservation manager use a prefix, which allows us to "namespace" these resources. The problem with this approach is that when metrics are produced for the prefixed resources, it make it increasingly more difficult to group by name when they are namespaces with a prefix.

Instead of using the prefixed names in the metrics, it would be better if a "namespace" dimension was used to help group the same resource names.

simitt:
@marclop the goal of this task is to remove  the `namespace` from a metric name and instead set it as dimension on the metric document, right?
Is it clear how the prefix can be identified?

marclop:
Correct that's the overall goal. The specifics are a bit unclear at the moment, but we'll probably need to surface it as a field in the configuration structs and propagate it down to the metrics.

PR #242, opened 2023-07-14
pubsublite: add resource namespacing

axw:
If CommonConfig.Namespace is non-empty, a prefix (filter) will be added to all namespaces, topics, and subscriptions.
The namespace will be stripped from returned names, and will be added as a separate attribute to metrics and logs.

Part of https://github.com/elastic/apm-queue/issues/224

### DIFF
--- a/pubsublite/consumer_test.go
+++ b/pubsublite/consumer_test.go
@@ -19,11 +19,20 @@ package pubsublite
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
+	"net"
 	"strings"
+	"sync"
 	"testing"
 
+	"cloud.google.com/go/pubsublite/apiv1/pubsublitepb"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
+	"google.golang.org/api/option"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
 
 	apmqueue "github.com/elastic/apm-queue"
 )
@@ -32,12 +41,12 @@ func TestNewConsumer(t *testing.T) {
 	t.Run("empty config", func(t *testing.T) {
 		_, err := NewConsumer(context.Background(), ConsumerConfig{})
 		assert.EqualError(t, err, "pubsublite: invalid consumer config: "+strings.Join([]string{
-			"pubsublite: project must be set",
-			"pubsublite: region must be set",
-			"pubsublite: logger must be set",
-			"pubsublite: at least one topic must be set",
-			"pubsublite: consumer name must be set",
-			"pubsublite: processor must be set",
+			"logger must be set",
+			"region must be set",
+			"project must be set",
+			"at least one topic must be set",
+			"consumer name must be set",
+			"processor must be set",
 		}, "\n"))
 	})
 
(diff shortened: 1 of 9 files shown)


pubsublite/consumer_test.go
```diff
@@ -60,6 +69,283 @@ func TestNewConsumer(t *testing.T) {
...
+	server, config := newConsumerService(t)
+	topic1SubscriptionPath := "projects/project/locations/region-1/subscriptions/name_space-topic1+consumer_name"
+	topic2SubscriptionPath := "projects/project/locations/region-1/subscriptions/name_space-topic2+consumer_name"
+
+	server.startSession = func(session *subscriberSession, _ *pubsublitepb.SeekRequest) {
+		switch session.subscription {
+		case topic1SubscriptionPath:
+		case topic2SubscriptionPath:
+		default:
+			panic("unexpected subscription: " + session.subscription)
```
marclop, line 92:
```suggestion
			t.Fatal("unexpected subscription: " + session.subscription)
```

pubsublite/consumer_test.go
```diff
@@ -60,6 +69,283 @@ func TestNewConsumer(t *testing.T) {
...
+		}
+		switch session.partition {
+		case 1, 2:
+		default:
+			panic(fmt.Errorf("unexpected partition: %d", session.partition))
```
marclop, line 97:
```suggestion
			t.Fatalf("unexpected partition: %d", session.partition)
```

marclop (approved):
LGTM and thanks for adding the subscriber and publisher serviceServers!

pubsublite/manager.go
```diff
@@ -101,13 +105,22 @@ func (m *Manager) ListReservations(ctx context.Context) ([]string, error) {
 			}
 			return nil, fmt.Errorf("pubsublite: failed listing reservations: %w", err)
 		}
-		names = append(names, path.Base(reservation.Name))
+		reservationName := path.Base(reservation.Name)
+		if !strings.HasPrefix(reservationName, namespacePrefix) {
+			continue
+		}
+		names = append(names, reservationName[len(namespacePrefix):])
```
endorama, line 112: Q: why are we returning names without prefix here?

pubsublite/manager.go
```diff
@@ -101,13 +105,22 @@ func (m *Manager) ListReservations(ctx context.Context) ([]string, error) {
...
 	}
 }
 
 // ListReservationTopics lists topics in the given reservation.
```
endorama, line 116:
```suggestion
// ListReservationTopics lists topics in the given reservation.
//
// If the CommonConfig.Namespace is non-empty, then only reservations topics 
// matching that namespace are returned; the returned reservation topic name
// does not include the namespace.
```

pubsublite/consumer_test.go
axw: This code is running in a different goroutine - t.Fatal should only be called from the main test goroutine.

pubsublite/manager.go
axw: The idea is that the namespace should only ever be specified in one place, and only ever surfaced as a separate dimension (in metrics) or attribute (in logs and traces). That way one can, for example, compare metrics for a topic with the same name in two different namespaces.

endorama: I see, I also read your updated comment. As PubSub does not support the concept of "namespace" we consider it an implementation detail so we hide it to callers. Thanks for the clarification.

PR #246, opened 2023-07-18
kafka: add topic namespacing

axw:
If `CommonConfig.Namespace` is non-empty, a prefix (filter) will be added to all topics. This enables the creation of a topic with the same logical name within two different namespaces. The namespace will be stripped from returned topic names, and will be added as a separate attribute to metrics and logs.

Closes https://github.com/elastic/apm-queue/issues/224

### DIFF
--- a/kafka/manager.go
+++ b/kafka/manager.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/twmb/franz-go/pkg/kadm"
 	"github.com/twmb/franz-go/pkg/kerr"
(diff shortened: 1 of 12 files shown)


endorama (approved):
I mainly compared these changes with #242 to ensure they share a similar behaviour. That's what my approval is about.

I didn't find any other major issue but I don't feel confident in the codebase enough to confirm the entire changeset is complete, so merging after another review would be preferred.

kafka/manager.go
```diff
@@ -106,18 +108,16 @@ func (m *Manager) DeleteTopics(ctx context.Context, topics ...apmqueue.Topic) er
...
 				logger.Debug("kafka topic does not exist")
 			} else {
 				span.RecordError(err)
 				span.SetStatus(codes.Error, "failed to delete one or more topic")
 				deleteErrors = append(deleteErrors,
-					fmt.Errorf(
-						"failed to delete topic %q: %w",
-						response.Topic, err,
-					),
+					fmt.Errorf("failed to delete topic %q: %w", topic, err),
```
marclop, line 120: TIL `%q`

marclop (approved):
Looks like a good change

Issue #139, opened 2023-05-15
discussion: expose buffered records thresold and timeout


kruskall:
Opening this issue for a potential discussion after some issues discovered in https://github.com/elastic/apm-queue/pull/116

We currently have no control on records buffering. The kafka producer has options to control the feature (maxbufferedrecords and producer linger) but they are not currently exposed in the library.

Should they be exposed ? Should we treat buffering as an implementation detail ? 

https://pkg.go.dev/github.com/twmb/franz-go@v1.13.3/pkg/kgo#ProducerLinger
https://pkg.go.dev/github.com/twmb/franz-go@v1.13.3/pkg/kgo#MaxBufferedRecords
https://pkg.go.dev/github.com/twmb/franz-go@v1.13.3/pkg/kgo#ManualFlushing

kruskall:
I've updated the title as the issue also applies to pubsub with similar options

marclop:
This may come into play when we start looking at #84.

PR #249, opened 2023-07-27
feat: expose max records and bytes options in kafka config

kruskall:
- Add `MaxPollBytes` and `MaxPollPartitionBytes` to consumer
- Add `MaxBufferedRecords`, `ProducerBatchMaxBytes` and `ManualFlushing` to producer

I think we should expose timeouts too but I'm not sure how to approach that. It would only concern the producer but I don't think we should expose all of them as it could get confusing. The available timeouts are:
- [ProduceRequestTimeout](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#ProduceRequestTimeout)
- [RecordDeliveryTimeout](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#RecordDeliveryTimeout)
- [TransactionTimeout](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#TransactionTimeout)

Closes https://github.com/elastic/apm-queue/issues/139


kafka/consumer.go
```diff
@@ -90,6 +96,12 @@ func (cfg *ConsumerConfig) finalize() error {
 	if cfg.Processor == nil {
 		errs = append(errs, errors.New("kafka: processor must be set"))
 	}
+	if cfg.MaxPollBytes < 0 {
+		errs = append(errs, errors.New("kafka: max pool bytes cannot be negative"))
```
dmathieu, line 100:
```suggestion
		errs = append(errs, errors.New("kafka: max poll bytes cannot be negative"))
```

kafka/consumer.go
```diff
@@ -90,6 +96,12 @@ func (cfg *ConsumerConfig) finalize() error {
...
+	}
+	if cfg.MaxPollPartitionBytes < 0 {
+		errs = append(errs, errors.New("kafka: max pool partition bytes cannot be negative"))
```
dmathieu, line 103:
```suggestion
		errs = append(errs, errors.New("kafka: max poll partition bytes cannot be negative"))
```

dmathieu (approved):

simitt:
@kruskall is anything blocking from merging this PR in?

Issue #84, opened 2023-04-11
Benchmarking: Collect throughput metrics for AWS MSK


marclop:
## Description

We want to test and benchmark producing and consuming against Kafka. The goal is to establish a throughput baseline and we'll use the cloud provider _AWS MSK_ for this. Since

We MUST collect throughput and usage metrics.

## Design

To keep things simple, we could start by creating a single command under `cmd/queuebench`. We may retrofit the functionality into `testing.B`, but it'll be quicker and less cumbersome to start with a single command that runs a single benchmark for some time and prints the results.

We'll run a single producer and consumer for the benchmark, collect the metrics using a Manual OTLP reader.

https://github.com/elastic/apm-queue/blob/63f6f01bf68a76aad5fd2f7bd44af0f7519f8e16/kafka/metrics_test.go#L323C2-L323C2

After the benchmark has finished, we'll collect the metrics that we want to print, perform any calculations (such as `events/s` and `bytes/s`).

### Setup

The idea is to run a the benchmarks against a single topic (for now) and a single partition (we can expose a `-partitions` flag which allows setting the number of partitions for the topic, but default to `1` for now). The command will:

1. Create a unique topic name with the specified number of partitions (`-partitions`).
2. Generate N events with the specified byte size (`-event-size`).
3. Create a manual reader for OTLP Exporter.
5. Create a consumer and start consuming in a goroutine.
6. While timer isn't exceeded, produce events.
7. Stop producer.
8. Stop consumer after all events have been consumed.
9. Collect metrics and perform any necessary calculations to print the results.

### Inputs

- `-event-size` in bytes, default to 1024 (1KB).
- `-partitions` default to 1. (keep it to a single topic producing for now).
- `-duration` the amount of time to run the benchmark for.

### Output (derived from metrics)

For now, we can just output the metrics in a table format. We may change this over time as we run this benchmarks periodically and in CI.

- Bytes/s
- Events/s
- Produce error rate in %.
- Consumption latency. We could print p50, p90, p99.

simitt:
@marclop @axw could you narrow the expectations for this task a bit more and add some definition of how the benchmark setup should look like.

endorama:
Adding here some questions:
- do we have tooling to generate data?
- I expect we are not yet collecting all the needed metrics for this benchmark, is that correct?
- Do we store those metrics long term?
- How should this benchmark system be monitored?

marclop:
[...]

PR #256, opened 2023-08-04
Collect throughput metric to benchmark AWS MSK

endorama:
Add `queuebench` command to perform controlled production to and consumption from an AWS MSK Serverless cluster.

The command:
- create a dedicated topic with a unique name
- produce events; event is a random bytes sequence of given length
- consume events
- print some metrics collected by otel
- delete created topic

Closes https://github.com/elastic/apm-queue/issues/84

AWS MSK cluster creation is not part of this PR.

Is easier to run benchmarks from an EC2 by uploading this command and running with appropriate configurations.

`queuebench` command has these command line flags:
- `brokers` (mandatory): the list of brokers bootstrap URLs to connect to
- `duration` (mandatory): the duration time, in seconds, of the benchmark run
the benchmark run
- `verbose`: a flag to enable more verbose debugging, in particular Zap debug logs for Kafka clients.

Two config options are not exposed yet through flags:
- `event-size` (default `1024`): the byte size of events being sent as records
- `partitions` (default `1`): the number of partitions to split the single topic created for


tools/queuebench/bench.go
axw, line 1: Can you please move this from `tools/` to `cmd/`? The `tools` directory is for build/CI-related tooling.

tools/queuebench/Taskfile.yml
axw, line 1: Did you intend to commit this?

tools/queuebench/go.mod
```diff
@@ -0,0 +1,48 @@
+module github.com/elastic/apm-queue/cmd/queuebench
```
axw, line 1: We used to have multiple modules -- we had one in systemtest and removed it: https://github.com/elastic/apm-queue/pull/214. We removed that one because it was causing issues with Renovate.

So with that said, can we avoid introducing this module? IIANM the module dependencies you have added that aren't in the main apm-queue module are `pflag` and `uitable`. I think we could get by without those for now to keep things simple as a whole. WDYT?

tools/queuebench/bench.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func createConsumer(commoncfg kafka.CommonConfig, topics []apmqueue.Topic) (*kafka.Consumer, error) {
+	commoncfg.ClientID = fmt.Sprintf("%s-consumer", app)
+
+	return kafka.NewConsumer(kafka.ConsumerConfig{
+		CommonConfig: commoncfg,
+
+		Topics:    topics,
+		GroupID:   "zero",
```
axw, line 144: Why "zero"?

tools/queuebench/main.go
```diff
@@ -0,0 +1,183 @@
...
+		if err != nil {
+			panic(err)
+		}
+
+		cancel()
+		log.Println("gracefully stopping consumer")
+		err = bench.Consumer.Close()
+		if err != nil {
+			panic(err)
+		}
```
axw, line 134: https://github.com/elastic/apm-queue/issues/84 says

> Stop consumer after all events have been consumed.

How do you know that all the events have been consumed?

axw (commented):
A few thoughts so far. I think it might be a good idea to do this in two steps - first implement the load generation & consumption, and then follow up with presenting metrics.

tools/queuebench/bench.go
endorama: I started in `cmd/` and moved to `tools/` as I though this was CI related :D 

But :+1:, I'll move it it.

tools/queuebench/Taskfile.yml
endorama: yes for the draft, as it showcase how I'm running it and tracking it in git helps with reproducing the test cases. The plan is to remove it or adapt it before this PR is ready for review.

tools/queuebench/go.mod
endorama: sure

tools/queuebench/bench.go
endorama: I didn't want to repeat "consumer" again and this is an uncommon value that I'm using to search for during testing. I haven't spent much time thinking about this as I was focused on the overall functionality. Open to suggestions for a better value.

endorama: I think I misunderstood some implications of `GroupID`. To prevent clients to try reading from past reached offsets due to belonging to the same group (as the cluster itself may not be recreated between runs) I changed it to be unique for the single run and time based, which should prevent conflicts.

See [`8ada4de` (#256)](https://github.com/elastic/apm-queue/pull/256/commits/8ada4ded9374bfea7a6e67dd6c2e7a6e23129eb4#diff-ab6dc1b362b05def35e2d65c689ac6234c2a73455016306e7af1ad54d94fc1ceR118)

tools/queuebench/main.go
endorama: Thanks to the refactoring the previous problematic behaviour is gone. I implemented two global counters to track produced and consumed ([`0931d0f` (#256)](https://github.com/elastic/apm-queue/pull/256/commits/0931d0fcaaf785e064cc6a754d5c8b30938cb5bc)) and a waiting loop that terminates on `consumed total = produced total` ([`6e8e189` (#256)](https://github.com/elastic/apm-queue/pull/256/commits/6e8e189ca8ee8bd473ae3738b7d5737f2ee0cd10))

endorama:
Based on different comment from @axw I moved forward with a new implementation from scratch.

Relevant changes from the previous implementation:
- the code is now located in `cmd/queuebench`
- no usage of other Go packages, no need for a different `go.mod`
- refactored production and consumption logic, using global counters to track consumer work and finish benchmark when consumer catch up with produced records
- calculate benchmark duration
- metric display is simplified, no table output
- display benchmark duration, total produced and consumed as reported by global counters and print Kafka metrics

There are still some TODOs to complete that I plan to address in the next commits.

tools/queuebench/bench.go
axw: The topic name is unique per run, right? So I don't think it's necessary to also make this ID unique. Anyway, looks better now, thanks.

tools/queuebench/main.go
axw: > At first I tried counting the number of produced messages, but then I didn't find a way to count consumed messages that I considered effective (apmqueue.Processor does not allow pointer receiver and I wasn't keen on using a global).

`apmqueue.Processor` is an interface, so you can definitely use a pointer receiver.

> I considered using otel metrics but I wasn't then sure how to account for possible consumption errors (how to stop consuming if the number of produced and consumed messages do not match for an unforeseen reason? Would a timeout be enough here?)

We could have the same issue (consumed never reaches produced) regardless of how the numbers are measured. Yes, I think having a timeout would be a good idea.

> Thanks to the refactoring the previous problematic behaviour is gone. I implemented two global counters to track produced and consumed ([0931d0f (#256)](https://github.com/elastic/apm-queue/pull/256/commits/0931d0fcaaf785e064cc6a754d5c8b30938cb5bc)) and a waiting loop that terminates on consumed total = produced total ([6e8e189 (#256)](https://github.com/elastic/apm-queue/pull/256/commits/6e8e189ca8ee8bd473ae3738b7d5737f2ee0cd10))

:+1:

cmd/queuebench/main.go
```diff
@@ -0,0 +1,159 @@
...
+
+	log.Println("start consumer")
+	go func() {
+		if err := bench.c.Run(ctx); err != nil {
+			log.Panicf("consumer run ended with an error: %s", err)
+		}
+	}()
+
+	log.Println("start producing")
+	produce(ctx, bench.p, bench.Topics[0], cfg.eventSize)
```
axw, line 107: need to check the error

cmd/queuebench/main.go
```diff
@@ -0,0 +1,159 @@
...
+	log.Println("stop producing")
+
+	for {
+		// TODO: add timeout
+		if totalconsumed < totalproduced {
+			time.Sleep(100 * time.Millisecond)
+		} else {
+			break
+		}
+	}
```
axw, line 118: Another option here could be to:
 - when the producer is done, tell the consumer how many messages were produced
 - have the consumer signal when it has consumed that many messages (e.g. by closing a channel)

Then this code can just block until that is signalled, or until the context is cancelled. This should minimise observer effect.

endorama: I worked on this implementation today but I have some doubts about it being the right solution considering the trade off with simplicity.

I refactored the implementation to use a ticker and react to a timeout (default `1m`) and I wondered if the a delay of at maximum 100ms is worth choosing the channel based implementation. We could also reduce the ticker time to a lesser value, the assumption on which I chose 100ms is I considered it an effective trade off: benchmarks will run for seconds if not minutes so a skew of < 99ms would not impact the overall result significantly.

My preferred approach here would be to leverage otel metrics emitted by producer and consumer (replacing the globals in use now), but this would not fit the channel based design.

cmd/queuebench/main.go
endorama: done in [1ec8dbf](https://github.com/elastic/apm-queue/pull/256/commits/1ec8dbfb865f99ff0a015a7d8666a323621d0c60)

cmd/queuebench/main.go
```diff
@@ -0,0 +1,209 @@
...
+	log.Println("production ended")
+
+	log.Println("waiting for consumer to fetch all records")
+	ticker := time.NewTicker(100 * time.Millisecond)
+	timer := time.NewTimer(cfg.timeout)
+wait:
+	for {
+		select {
+		case <-timer.C:
+			ticker.Stop()
```
axw, line 145:
```suggestion
```

cmd/queuebench/main.go
```diff
@@ -0,0 +1,209 @@
...
+			log.Println("reached timeout, moving on")
+			break wait
+		case <-ticker.C:
+			if totalconsumed >= totalproduced {
+				log.Println("consumption ended")
+				ticker.Stop()
```
axw, line 151:
```suggestion
```

cmd/queuebench/main.go
```diff
@@ -0,0 +1,209 @@
...
+				break wait
+			}
+		}
+	}
+	timer.Stop()
```
axw, line 156:
```suggestion
```

cmd/queuebench/main.go
```diff
@@ -0,0 +1,209 @@
...
+	buf := make([]byte, size)
+
+	_, err := rand.Read(buf)
+	if err != nil {
+		return fmt.Errorf("cannot read random bytes: %w", err)
+	}
+
+	record := apmqueue.Record{
+		Topic: topic,
+		// OrderingKey: []byte{},
```
axw, line 196:
```suggestion
```

cmd/queuebench/main.go
axw: I was thinking more about CPU consumption that polling adds. I'm fine with using the OTel metrics for this if the overhead of gathering the metrics every 100ms isn't significant. The current implementation isn't concurrent-safe, as when there are multiple partitions there may be multiple consumer goroutines. Relying on either a channel approach or OTel metrics should resolve that.

axw (commented):
I think this is looking good, thanks. Let me know when this is ready for proper review.

cmd/queuebench/main.go
```diff
@@ -0,0 +1,209 @@
...
+	if err := produce(ctx, bench.p, bench.Topics[0], cfg.eventSize, cfg.duration); err != nil {
+		log.Panicf("error while producing records: %s", err)
+	}
+	productionduration := time.Since(productionstart)
+
+	log.Println("production ended")
+
+	log.Println("waiting for consumer to fetch all records")
+	ticker := time.NewTicker(100 * time.Millisecond)
+	timer := time.NewTimer(cfg.timeout)
```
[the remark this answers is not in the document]
endorama: I haven't used defer in this case as I wanted to stop both in the proximity of the for loop instead of at the end of `main`, but doesn't make much of a difference and `defer` is more idiomatic so :+1:

cmd/queuebench/main.go
endorama: > I was thinking more about CPU consumption that polling adds. I'm fine with using the OTel metrics for this if the overhead of gathering the metrics every 100ms isn't significant. 

I will test and report back. One question about the CPU consumption: I was under the assumption that we aim to establish a baseline, not to make a realistic estimation of throughput. The implication to me is that as long as conditions are consistent across benchmark we shouldn't worry too much on ensuring the benchmark code performs as its best.
What's your thoughts on this?

> The current implementation isn't concurrent-safe, as when there are multiple partitions there may be multiple consumer goroutines. Relying on either a channel approach or OTel metrics should resolve that.

Indeed, I'll investigate the CPU consumption for using otel metrics and decide which approach to choose base don it.

endorama: I profiled the code and my conclusion is that the impact is not significant, but I'd like to review the results tomorrow with @axw

axw: > One question about the CPU consumption: I was under the assumption that we aim to establish a baseline, not to make a realistic estimation of throughput. The implication to me is that as long as conditions are consistent across benchmark we shouldn't worry too much on ensuring the benchmark code performs as its best.
What's your thoughts on this?

It's not so much about comparing one run to another, but about identifying the maximum throughput so we compare against users of apm-queue. It can be approximate, but the closer to the true maximum the better.

endorama: We reviewed the profiling data and confirmed that using otel metrics does not impact the benchmark.

cmd/queuebench/main.go
```diff
@@ -0,0 +1,209 @@
...
+	defer timer.Stop()
+
+	var rm metricdata.ResourceMetrics
+	totalproduced := int64(0)
+	totalconsumed := int64(0)
+wait:
+	for {
+		ctx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
+		defer cancel()
```
axw, line 145:
```suggestion
```

I guess the idea here is to ensure the collection doesn't take longer than the interval? I think it's probably unnecessary, but if you want to keep it then the defer needs fixing - the defer won't fire until the main function ends, and will accumulate for every iteration of the loop.

cmd/queuebench/main.go
```diff
@@ -0,0 +1,209 @@
...
+		select {
+		case <-timer.C:
+			log.Println("reached timeout, moving on")
+			break wait
+		case <-ticker.C:
+			rdr.Collect(ctx, &rm)
```
axw, line 151: Should we be checking the result here? Otherwise how would we react to the context being cancelled?

cmd/queuebench/main.go
```diff
@@ -0,0 +1,209 @@
...
+
+	}
+	defer teardown()
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		for range c {
+			teardown()
+		}
+	}()
```
axw, line 109:
```suggestion
	ctx, stop := signal.NotifyContext(ctx, os.Interrupt)
	defer stop()
```

If we cancel the context on interrupt, then we can handle context cancellation to break the producing/consuming/wait loop. Then teardown can happen in one place.

axw (approved):
LGTM, just a few minor comments and suggestions about how teardown/cancellation is done.

cmd/queuebench/main.go
endorama: otel code documentation says:
> 	  // Collect gathers and returns all metric data related to the Reader from
>	  // the SDK and stores it in out. An error is returned if this is called
>	  // after Shutdown or if out is nil.
>	  Collect(ctx context.Context, rm *metricdata.ResourceMetrics) error

I don't think is possible to call it after `Shutdown` in this case, but I added error handling for the `nil` case.

I added a select case to intercept `ctx.Done` to detect context cancellation.

cmd/queuebench/main.go
endorama: yes, that was the idea, and the defer comes from an initial implementation where it was within a function :( Removed as I too don't think it's necessary.

cmd/queuebench/main.go
endorama: I refactored the logic based on your suggestion in [e1a0be2](https://github.com/elastic/apm-queue/pull/256/commits/e1a0be2a31c4166ba0cc9617da85e8752e550d4b)

cmd/queuebench/main.go
```diff
@@ -200,6 +200,11 @@ func produce(ctx context.Context, p *kafka.Producer, topic apmqueue.Topic, size
 
 	deadline := time.Now().Add(duration)
 	for time.Now().Before(deadline) {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
```
axw, line 216: Is this necessary? Does the Produce method below not honour the context cancellation?

cmd/queuebench/main.go
```diff
@@ -0,0 +1,223 @@
...
+	log.Println("bench run completed successfully")
+}
+
+func produce(ctx context.Context, p *kafka.Producer, topic apmqueue.Topic, size int, duration time.Duration) error {
+	buf := make([]byte, size)
+
+	_, err := rand.Read(buf)
+	if err != nil {
+		return fmt.Errorf("cannot read random bytes: %w", err)
+	}
```
axw, line 203:
```suggestion
	if _, err := rand.Read(buf); err != nil {
		return fmt.Errorf("cannot read random bytes: %w", err)
	}
```

cmd/queuebench/main.go
```diff
@@ -0,0 +1,223 @@
...
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+		if err = p.Produce(ctx, record); err != nil {
+			return err
+		}
+	}
+
+	return err
```
axw, line 222:
```suggestion
		if err := p.Produce(ctx, record); err != nil {
			return err
		}
	}

	return nil
```

The error will always be nil at the end of the loop. By keeping the `err` variables minimally scoped, we can prove that is the case, since then there will be no `err` in scope at the end of the loop.

axw (approved):

cmd/queuebench/main.go
endorama: The `Produce` method does not but context cancellation is handled by `franz-go/kgo` (see [docs](https://github.com/twmb/franz-go/blob/89cd6a7c4fef29a400bba5a6cd8a77def5ec8105/pkg/kgo/producer.go#L326-L356)), I removed it.

cmd/queuebench/main.go
endorama: added in [edac17a](https://github.com/elastic/apm-queue/pull/256/commits/edac17a5bff07217794dfc87f91b9ad23c009f5e)

cmd/queuebench/main.go
endorama: updated in [edac17a](https://github.com/elastic/apm-queue/pull/256/commits/edac17a5bff07217794dfc87f91b9ad23c009f5e)

Issue #251, opened 2023-07-31
Update Kafka producer metric names


marclop:
## Description

Create a new metric that encompasses the count of messages produced and use a dimension to report the produce outcome. The new metric name must be `producer.messages.count` and use two dimensions, `outcome=success` and `outcome=failure` to report whether a record was produced or not.

Remove the two metrics below since the new metric will cover both outcomes.

https://github.com/elastic/apm-queue/blob/63f6f01bf68a76aad5fd2f7bd44af0f7519f8e16/kafka/metrics.go#L39-L40

endorama:
The relevant PR in apm-queue has been merged. I'll proceed in updating the related dashboards and alerts.

endorama:
Dashboards PR merged for all environments. Waiting to confirm the sync process worked correctly.

endorama:
Dashboard updates confirmed in staging and production. Closing.

@marclop @kyungeunni I noticed there are some cloned dashboard with the fixes applied, I think they can be removed now.

PR #287, opened 2023-09-27
kafka: change metrics (producer.messages.count)

endorama:
Add Kafka `producer.messages.count` metric. This metric counts all produced
messages and uses `outcome=success|failure` dimension to distinguish
between messages that failed or have been successfully produced.

`error=other|timeout|cancelled` dimension is used to identify the failure
reason.

Remove `producer.messages.produced` and `producer.messages.errored`
Kafka metrics as both are replaced by `producer.messages.count`.

Part of https://github.com/elastic/apm-queue/issues/251

### DIFF
--- a/kafka/metrics.go
+++ b/kafka/metrics.go
@@ -18,8 +18,6 @@
 package kafka
 
 import (
-	"context"
-	"errors"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


marclop:
@endorama, would you mind reopening the PR from a branch in this repository? PRs from forks skip the system tests.

kafka/metrics.go
```diff
@@ -111,20 +102,24 @@ func (h *metricHooks) OnProduceRecordUnbuffered(r *kgo.Record, err error) {
 	}
 
 	if err != nil {
-		errorType := attribute.String("error", "other")
+		errorTypeAttr := attribute.String("error", "other")
```
marclop, line 105: I just noticed a mismatch between this field name here and the spec we outlined; it should be `error_reason` per the spec. Perhaps instead of "other," it could be `"unknown"`

endorama: yes, I was planning to address this in a separate PR

PR #289, opened 2023-09-28
kafka: change metrics (producer.messages.count)

endorama:
Add Kafka `producer.messages.count` metric. This metric counts all produced
messages and uses `outcome=success|failure` dimension to distinguish
between messages that failed or have been successfully produced.

`error=other|timeout|cancelled` dimension is used to identify the failure
reason.

Remove `producer.messages.produced` and `producer.messages.errored`
Kafka metrics as both are replaced by `producer.messages.count`.

Part of https://github.com/elastic/apm-queue/issues/251

PR #290, opened 2023-09-28
kafka: use error_reason dimension

endorama:
This PR aligns `kafka` package to use `error_reason` dimension instead of the more generic `error`.

It also changes the `other` error reason to `unknown`, which better represent the reason.

Requires https://github.com/elastic/apm-queue/pull/289


dmathieu (dismissed):

simitt:
@endorama is there anything holding back from merging this in?

PR #297, opened 2023-10-10
test: refactor unit tests to reduce flakyness

kruskall:
Tested locally by running tests with `-count=<big>` and was able to reduce flakyness significantly.

### TestConsumerDelivery:

Instead of waiting for `consumer.Close` to complete before proceeding, call `consumer.Close` from the same goroutine and wait for `consumer.Run` to exit. This ensure that the first consumer has exited.

Call the `cancel` function inside the processor to avoid a small race  between the time the processor finishes and the time `cancel` was called.

### TestNewProducerBasic

Set `MaxBufferedRecords` to 0 to avoid context cancelled because the consumer is waiting on the buffered records. (another option would be to close the producer to force a flush).

### DIFF
--- a/kafka/consumer_test.go
+++ b/kafka/consumer_test.go
@@ -281,8 +281,9 @@ func TestConsumerDelivery(t *testing.T) {
 
 			var processed atomic.Int32
 			var errored atomic.Int32
-			newProcessor := func(processRecord, failRecord <-chan struct{}) apmqueue.Processor {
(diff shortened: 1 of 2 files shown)


endorama (approved):
Running tests with `count=50` and `count=100` shows already a great difference between `main` and this branch.
I observed 0 failure for this branch and multiple failures for `main`, so :+1: from me.

PR #299, opened 2023-10-11
kakfa: update existing topics

axw:
Updates `CreateTopics` to update the number of partitions, and alter the
topic configuration when the topic already exists.


axw (commented):
Unit test looks good. Should we also add something to the system tests?

kafka/topiccreator.go
```diff
@@ -116,36 +116,102 @@ func (c *TopicCreator) CreateTopics(ctx context.Context, topics ...apmqueue.Topi
...
+			logger.Info("updated partitions for kafka topic",
+				zap.String("topic", topicName),
+			)
+		}
+		if len(c.topicConfigs) > 0 {
+			alterCfg := make([]kadm.AlterConfig, 0, len(c.topicConfigs))
+			for k, v := range c.topicConfigs {
+				alterCfg = append(alterCfg, kadm.AlterConfig{Name: k, Value: v})
+			}
+			alterResp, err := c.m.adminClient.AlterTopicConfigs(ctx, alterCfg,
```
lahsivjar, line 188: (not for this PR) What do you think about caching the successfully applied configs and comparing them with latest configs (for both topic configs and partition count) to avoid making this call if no changes have been made?

lahsivjar (dismissed):
LGTM! One comment and one discussion point.

kafka/topiccreator.go
marclop: We could open an issue, although I don't think it matters much in practice.

kafka/topiccreator.go
```diff
@@ -116,36 +116,102 @@ func (c *TopicCreator) CreateTopics(ctx context.Context, topics ...apmqueue.Topi
...
 	}
-	if c.origTopicConfigs != nil {
+	if len(c.origTopicConfigs) > 0 {
 		createTopicParamsFields = append(createTopicParamsFields,
 			zap.Reflect("topic_configs", c.origTopicConfigs),
 		)
 	}
-	var createErrors []error
+
+	var existingTopics []string
```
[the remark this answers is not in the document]
marclop: Thanks for noticing! Done.

kafka/topiccreator.go
lahsivjar: Hmm fair point, does seem like too much effort. Even if there are rate limits in the admin client we can probably tweak our period of reconciliation.

lahsivjar (approved):

PR #302, opened 2023-10-11
kafka: Remove record-level tracing

### LINKED ISSUE #300: telemetry: don't trace on a per-event basis
We've been doing some performance testing and creating a trace on a per-event basis absolutely destroys performance by using a lot of memory. We should still trace on each of the Kafka read/write, but not on a per-event basis

marclop:
To reduce memory usage of applications that use this library, remove the usage of `kotel` tracing hooks to avoid creating a span for each event that is either read or written to Kafka.

Closes #300


endorama (approved):
looks good to me. In #300 you mention we want to trace kafka read/write, but I expect it will be part of a follow-up PR

lahsivjar (approved):

marclop:
@endorama exactly

PR #316, opened 2023-10-30
kafka: manager: List topics before creating them

marclop:
Updates the manager `CreateTopics` method to only issue the CreateTopics call if the topics don't exist in Kafka, preventing undesirable errors from being returned and logged.

### DIFF
--- a/kafka/topiccreator.go
+++ b/kafka/topiccreator.go
@@ -101,30 +101,51 @@ func (c *TopicCreator) CreateTopics(ctx context.Context, topics ...apmqueue.Topi
 	for i, topic := range topics {
 		topicNames[i] = fmt.Sprintf("%s%s", namespacePrefix, topic)
 	}
-	responses, err := c.m.adminClient.CreateTopics(
-		ctx,
+
+	existing, err := c.m.adminClient.ListTopics(ctx)
+	if err != nil {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
+		return fmt.Errorf("failed to list kafka topics: %w", err)
+	}
+
+	// Build two lists, one for the topics that haven't been created yet, and
+	// another one that can be used to update a topic's partitions.
+	missingTopics := make([]string, 0, len(topicNames))
+	updatePartitions := make([]string, 0, len(topicNames))
+	for _, wantTopic := range topicNames {
+		if !existing.Has(wantTopic) {
+			missingTopics = append(missingTopics, wantTopic)
+			continue
+		}
+		if len(existing[wantTopic].Partitions) < c.partitionCount {
+			updatePartitions = append(updatePartitions, wantTopic)
+		}
+	}
+	fmt.Printf("%+v\n", missingTopics)
+
+	responses, err := c.m.adminClient.CreateTopics(ctx,
 		int32(c.partitionCount),
 		-1, // default.replication.factor
 		c.topicConfigs,
-		topicNames...,
+		missingTopics...,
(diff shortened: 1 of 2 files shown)


kafka/topiccreator.go
```diff
@@ -101,30 +101,51 @@ func (c *TopicCreator) CreateTopics(ctx context.Context, topics ...apmqueue.Topi
...
+	for _, wantTopic := range topicNames {
+		if !existing.Has(wantTopic) {
+			missingTopics = append(missingTopics, wantTopic)
+			continue
+		}
+		if len(existing[wantTopic].Partitions) < c.partitionCount {
+			updatePartitions = append(updatePartitions, wantTopic)
+		}
+	}
+	fmt.Printf("%+v\n", missingTopics)
```
axw, line 125:
```suggestion
```

kafka/topiccreator.go
```diff
@@ -180,37 +204,37 @@ func (c *TopicCreator) CreateTopics(ctx context.Context, topics ...apmqueue.Topi
...
-			alterResp, err := c.m.adminClient.AlterTopicConfigs(ctx, alterCfg,
-				existingTopics...,
+	}
+	if len(c.topicConfigs) > 0 {
+		alterCfg := make([]kadm.AlterConfig, 0, len(c.topicConfigs))
+		for k, v := range c.topicConfigs {
+			alterCfg = append(alterCfg, kadm.AlterConfig{Name: k, Value: v})
+		}
+		alterResp, err := c.m.adminClient.AlterTopicConfigs(ctx,
+			alterCfg, topicNames...,
```
axw, line 214: Should we only be altering topic configs for topics that we didn't just create? We'll get a lot of log spam otherwise won't we?

marclop: We could, I didn't as to decouple the two... But I suppose we can. I can change it if you think it's significantly better. If it's just due to the logging, we could consolidate the info logging to a single line and use a topic slice instead of a single log line per topic altered.

axw: Even if it's one log line vs. N, I'd rather not see an Info log every time this function is called if it's effectively a no-op. That aside, it would be good to minimise the number of network calls - that's a smaller issue though, considering how infrequently this should be called.

marclop: I've updated the code and the test a bit in 8acf0af. Let me know what you think.

kafka/topiccreator.go
```diff
@@ -180,37 +206,37 @@ func (c *TopicCreator) CreateTopics(ctx context.Context, topics ...apmqueue.Topi
...
 		}
-		if len(c.topicConfigs) > 0 {
-			alterCfg := make([]kadm.AlterConfig, 0, len(c.topicConfigs))
-			for k, v := range c.topicConfigs {
-				alterCfg = append(alterCfg, kadm.AlterConfig{Name: k, Value: v})
-			}
-			alterResp, err := c.m.adminClient.AlterTopicConfigs(ctx, alterCfg,
-				existingTopics...,
+	}
+	if len(c.topicConfigs) > 0 {
```
axw, line 210:
```suggestion
	if len(existingTopics) > 0 && len(c.topicConfigs) > 0 {
```

axw (approved):
Looks good, thanks!

axw (dismissed):

marclop:
I've opened https://github.com/elastic/apm-queue/issues/318 to investigate.

axw (approved):

PR #319, opened 2023-11-07
feat: add tracing to kafka health functions

kruskall:


marclop (commented):
I'm unconvinced this is the right approach. This will add tracing to every Kafka ping, which isn't very useful.

PR #346, opened 2024-01-18
Kafka: Allow regex expression for topic monitoring

marclop:
Updates `kafka.Manager.MonitorConsumerLag` to allow monitoring unknown topics that match a regex expression.

### DIFF
--- a/kafka/manager.go
+++ b/kafka/manager.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/twmb/franz-go/pkg/kadm"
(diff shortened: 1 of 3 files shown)


queue.go
```diff
@@ -103,7 +103,12 @@ func (f ProcessorFunc) Process(ctx context.Context, rs ...Record) error {
 // Topic represents a destination topic where to produce a message/record.
 type Topic string
 
+// TopicConsumer is used to monitor a set of consumer topics.
 type TopicConsumer struct {
-	Topic    Topic
+	// Optional topic to monitor.
+	Topic Topic
```
axw, line 109: Do we want to keep this, or could we (eventually) just have the regex and make it required? It appears that Google Cloud Monitoring also supports regex matches: https://cloud.google.com/monitoring/api/v3/filters#filter_syntax

marclop: I expect we'll remove it later once we have no use for topics with known names. Until then, we need to keep it to maintain the current creation and monitoring mechanics.

kafka/manager.go
```diff
@@ -197,11 +216,18 @@ func (m *Manager) MonitorConsumerLag(topicConsumers []apmqueue.TopicConsumer) (m
 				}
 				topic = topic[len(namespacePrefix):]
 
+				var matchesRegex bool
+				for _, re := range regex {
+					if l.Group == re.consumer && re.regex.MatchString(topic) {
```
inge4pres, line 221: For educating myself: why do we need the Group to equal the consumer?

inge4pres (approved):
LGTM 👍🏼 
Due to being novel to the codebase I had to re-read a couple of times the tests and changes in `manager.go` but ultimately figure it out.
Left a question out of ignorance.

kafka/manager.go
marclop: The consumer is the consumer group. They're just named differently in the config vs the Kafka sdk

PR #368, opened 2024-02-08
kafka: Increase default `MaxPollRecords` to `500`

marclop:
Increases the default `MaxPollRecords` to `500`, which is inline with Kafka defaults.

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -55,7 +55,7 @@ type ConsumerConfig struct {
 	// events from Kafka brokers, it merely affects the number of records that
 	// are returned on `client.PollRecords`.
 	// The higher this setting, the higher the general processing throughput
-	// be. Gowever, when Delivery is set to AtMostOnce, the higher this number,
+	// be. However, when Delivery is set to AtMostOnce, the higher this number,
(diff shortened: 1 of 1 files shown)


kafka/consumer.go
```diff
@@ -50,11 +50,17 @@ type ConsumerConfig struct {
 	// GroupID to join as part of the consumer group.
 	GroupID string
 	// MaxPollRecords defines an upper bound to the number of records that can
-	// be polled on a single fetch. If MaxPollRecords <= 0, defaults to 100.
+	// be polled on a single fetch. If MaxPollRecords <= 0, defaults to 500.
+	// Note that this setting doesn't change how `franz-go` fetches and buffers
+	// events from Kafka brokers, it merely affects the number of records that
+	// are returned on `client.PollRecords`.
+	// The higher this setting, the higher the general processing throughput
+	// be. Gowever, when Delivery is set to AtMostOnce, the higher this number,
```
vikmik, line 58:
```suggestion
	// be. However, when Delivery is set to AtMostOnce, the higher this number,
```

vikmik (dismissed):
Thanks for the research!

vikmik (approved):

PR #399, opened 2024-03-05
kafka: Remove partition consumer goroutines

marclop:
Removes the need for dedicated goroutines to be spun when the consumer receives its partition assignments. Instead, it uses an `errgroup.Group` with a limit of `1`, which behaves like before.

This approach's primary advantage is that it reduces the number of running goroutines, which is noticeable as the partition count increases.

Without this patch, the consumer throughput plateaued after reaching a certain throughput.

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -28,6 +28,7 @@ import (
 	"github.com/twmb/franz-go/pkg/kgo"
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
 
 	apmqueue "github.com/elastic/apm-queue"
 	"github.com/elastic/apm-queue/queuecontext"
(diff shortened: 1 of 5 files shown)


kafka/consumer.go
```diff
@@ -450,10 +476,15 @@ func (c *consumer) processFetch(ctx context.Context, fetches kgo.Fetches) {
...
 			return
 		}
+		done := make(chan struct{})
+		go func() {
+			defer close(done)
+			consumer.consumeRecords(ftp)
+		}()
 		var err error
 		select {
 		// Send partition records to be processed by its dedicated goroutine.
```
carsonip, line 486: I wonder if "dedicated goroutine" still applies here. They are not goroutines readily waiting for records now.

carsonip (approved):
Skimmed through, looks great!

kafka/consumer.go
```diff
@@ -528,20 +581,27 @@ func (c *consumer) consumeTopicPartition(
 			}
 			last = i
 		}
-		// Only commit the records when at least a record has been processed when
-		// AtLeastOnceDeliveryType is set.
+		// Commit the last record offset when one or more records are processed
+		// and the delivery guarantee is set to AtLeastOnceDeliveryType.
+		// NOTE(marclop) When MaxProcessorConcurrency is greater than 1, commits
+		// may happen out of order, which would rewind the consumer. This should
+		// be fixed when AtLeastOnceDelivery is ready to be used.
```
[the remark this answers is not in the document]
axw: I don't think I understand the change or what needs fixing - could someone please clarify?
Isn't ALOD expected to rewind in case there are errors processing some data? What's new?

kafka/consumer.go
marclop: Good catch, I'll update the comment.

kafka/consumer.go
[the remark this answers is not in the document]
marclop: After deliberating with my pillow and some more testing, the `MaxProcessorConcurrency` flag doesn't appear to impact throughput. So I've removed the config flag and hardcoded the `errgroup.Group` limit to 1.

This means there's no perceivable change in execution from the event processing PoV, so this is a backwards-compatible change.

kafka/consumer.go
marclop: Fixed in the last commit.

lahsivjar (approved):
LGTM!

PR #418, opened 2024-03-22
feat: stop creating spans for every produced records

kruskall:
stop creating spans for every produced records


vikmik (approved):
(context: https://github.com/elastic/apm-ingest-service/issues/607 )

lahsivjar (approved):

PR #422, opened 2024-03-28
feat: use oidc for auth with GCP

amannocci:
## What is the change being made?

* Use OIDC to auth with Google services.

## Why is the change being made?

* Avoid the use of Vault and rely on keyless auth for CI.


marclop (approved):

v1v (commented):
🤟

PR #425, opened 2024-04-01
github-action: update condition

v1v:
### What

Support for push to main in addition to PRs that are not created by dependabot in the upstream.

### Why

https://github.com/elastic/apm-queue/pull/372 did solve the dependabot issue but it added a regression with the `main` branch and system tests were not triggered:

image

### DIFF
--- a/.github/workflows/system-test.yml
+++ b/.github/workflows/system-test.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+      - test-github
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/system-test.yml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - main
+      - test-github
```
v1v, line 8:
```suggestion
```

.github/workflows/system-test.yml
```diff
@@ -18,8 +19,8 @@ jobs:
       # This step is skipped if the pull request is from a forked repository or dependabot
       # In that case the job just creates a green status check on the pull request.
       - if: |
-          github.event.pull_request.head.repo.full_name == github.repository
-          && github.actor != 'dependabot[bot]'
+          ${{ ( github.event_name == 'push' && ( github.ref == 'main' || github.ref == 'test-github' ) ) ||
```
v1v, line 22:
```suggestion
          ${{ ( github.event_name == 'push' && github.ref == 'main'  ) ||
```

PR #434, opened 2024-04-09
Producer: add listener for bytes written to topic

vikmik:
This change allows us to record the number of bytes written to a topic
This PR adds a generic listener in the form of a `kgo.HookProduceBatchWritten`. The code that creates the producer may pass an optional listener function to be notified of writes to the topic.


vikmik:
Looks like the tests are already failing in `main`. Not sure what this is about yet

kafka/producer_test.go
```diff
@@ -146,7 +154,8 @@ func TestNewProducerBasic(t *testing.T) {
 			var actual []apmqueue.Record
 			orderingKeyToPartitionM := make(map[string]int32)
 			client.AddConsumeTopics(namespacedTopic)
-			for i := 0; i < len(batch)*partitionCount && len(actual) < len(batch); i++ {
```
vikmik, line 149: Unless I'm missing something, that first condition seemed incorrect ; if we want to read all records we must ensure that `len(actual) < len(batch)`.

kafka/producer.go
```diff
@@ -81,8 +81,20 @@ type ProducerConfig struct {
 	// If $KAFKA_PRODUCER_COMPRESSION_CODEC is not specified, then
 	// the default behaviour of franz-go is to use [snappy, none].
 	CompressionCodec []CompressionCodec
+
+	// BatchWriteListener specifies a callback function that will notify the caller of writes.
```
marclop, line 85:
```suggestion
	// BatchWriteListener is called per topic/partition after a batch is
	// successfully produced to a Kafka broker.
```

kafka/producer.go
```diff
@@ -81,8 +81,20 @@ type ProducerConfig struct {
...
+	BatchListener BatchWriteListener
+}
+
+type BatchWriteListener func(topic string, bytesWritten int)
+
+func (l BatchWriteListener) OnProduceBatchWritten(_ kgo.BrokerMetadata,
```
marclop, line 91: Would you mind adding a GoDoc comment for the type and public method?

kafka/producer_test.go
```diff
@@ -198,6 +207,13 @@ func TestNewProducerBasic(t *testing.T) {
 			//lint:ignore SA1012 passing a nil context is a valid use for this call.
 			fetches := client.PollRecords(nil, 1)
 			assert.Len(t, fetches.Records(), 0)
+
+			// Ensure that we recorded bytes written to the topic.
+			mut.Lock()
+			require.Len(t, topicBytesWritten, 1)
+			require.Contains(t, topicBytesWritten, "name_space-default-topic")
+			require.GreaterOrEqual(t, topicBytesWritten["name_space-default-topic"], 100)
```
marclop, line 215: Shouldn't this be `.Equal`? In which cases will this be Greater?

kafka/producer.go
```diff
@@ -81,8 +81,20 @@ type ProducerConfig struct {
 	// If $KAFKA_PRODUCER_COMPRESSION_CODEC is not specified, then
 	// the default behaviour of franz-go is to use [snappy, none].
 	CompressionCodec []CompressionCodec
+
+	// BatchWriteListener specifies a callback function that will notify the caller of writes.
+	BatchListener BatchWriteListener
+}
+
+type BatchWriteListener func(topic string, bytesWritten int)
```
marclop, line 89: I like how you encapsulated a subset of the behavior in a new type.

marclop (commented):
Change LGTM. Just some stylistic comments and a question on the test assertion.

kafka/producer_test.go
vikmik: This is slightly non-deterministic - the number of actual topic writes may vary, and hence the compressed bytes may vary.
I should probably add a comment for this

marclop (approved):
Thanks for addressing my comments. LGTM!

PR #447, opened 2024-04-24
feat: partition id as record metadata

inge4pres:
### Reason for this PR

We want to allow consumers to understand how many partitions they are poling from.
The partition Id is avaiable in both kafka and pubsublite so we should expose it and let consumers use it at their own advantage.

### Details

~We create a new un-exported string constant, with value `partition_id`. This will be populated in the context.Context attached to each kgo.Record via the `queuecontext` package.~
EDIT: accepted proposal to add the partition as a field of the `Record` struct.

Consumers will be ale to understand which partition a record was polled from by parsing the record field.

Unrelated small performance improvement: pre-allocate map capacity to the length of headers.

### DIFF
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -522,6 +522,10 @@ func newPartitionConsumer(ctx context.Context,
 	return &c
 }
 
+// This key is part of metadata as a mean to convey information
+// about the partition IDs that are assigned to each apmqueue.Processor.
+const metadataCtxPartitionKey = "partition_id"
+
 // consumeTopicPartition processes the records for a topic and partition. The
 // records will be processed asynchronously.
 func (c *pc) consumeRecords(ftp kgo.FetchTopicPartition) {
@@ -530,10 +534,14 @@ func (c *pc) consumeRecords(ftp kgo.FetchTopicPartition) {
 		// only the first record is received.
 		last := -1
 		for i, msg := range ftp.Records {
-			meta := make(map[string]string)
+			meta := make(map[string]string, len(msg.Headers))
 			for _, h := range msg.Headers {
 				meta[h.Key] = string(h.Value)
 			}
+			// Add the partition ID to the metadata to allow the processor to
+			// parallelize the processing of records from different partitions.
+			meta[metadataCtxPartitionKey] = fmt.Sprint(ftp.Partition)
+
 			processCtx := queuecontext.WithMetadata(msg.Context, meta)
 			record := apmqueue.Record{
 				Topic:       c.topic,
--- a/kafka/consumer.go
+++ b/kafka/consumer.go
@@ -522,10 +522,6 @@ func newPartitionConsumer(ctx context.Context,
 	return &c
 }
 
-// This key is part of metadata as a mean to convey information
-// about the partition IDs that are assigned to each apmqueue.Processor.
-const metadataCtxPartitionKey = "partition_id"
-
 // consumeTopicPartition processes the records for a topic and partition. The
 // records will be processed asynchronously.
 func (c *pc) consumeRecords(ftp kgo.FetchTopicPartition) {
@@ -538,13 +534,11 @@ func (c *pc) consumeRecords(ftp kgo.FetchTopicPartition) {
 			for _, h := range msg.Headers {
 				meta[h.Key] = string(h.Value)
 			}
-			// Add the partition ID to the metadata to allow the processor to
-			// parallelize the processing of records from different partitions.
-			meta[metadataCtxPartitionKey] = fmt.Sprint(ftp.Partition)
 
 			processCtx := queuecontext.WithMetadata(msg.Context, meta)
 			record := apmqueue.Record{
 				Topic:       c.topic,
+				Partition:   msg.Partition,
 				OrderingKey: msg.Key,
 				Value:       msg.Value,
 			}
--- a/kafka/consumer_test.go
+++ b/kafka/consumer_test.go
@@ -481,8 +481,9 @@ func TestConsumerContextPropagation(t *testing.T) {
 	}
 	processed := make(chan struct{})
 	expectedMeta := map[string]string{
-		"key":       "value",
-		"timestamp": time.Now().Format(time.RFC3339),
+		"key":          "value",
+		"timestamp":    time.Now().Format(time.RFC3339),
+		"partition_id": "0",
 	}
 	consumer := newConsumer(t, ConsumerConfig{
 		CommonConfig: commonCfg,
--- a/kafka/consumer_test.go
+++ b/kafka/consumer_test.go
@@ -139,6 +139,7 @@ func TestConsumerHealth(t *testing.T) {
 }
 
 func TestConsumerInstrumentation(t *testing.T) {
+	const partitions = 4
 	exp := tracetest.NewInMemoryExporter()
 	tp := sdktrace.NewTracerProvider(
 		sdktrace.WithSyncer(exp),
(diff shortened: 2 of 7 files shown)


lahsivjar:
> Partition id is not a concept tackled in other queues implementations managed here

CMIIW, we only have 2 implementations, kafka and pubsublite. For kafka, we can use franz-go's `OnPartition{Assigned, Lost}` ([ref](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#OnPartitionsAssigned)) methods and for pubsublite we can use the `ReassignmentHandler` ([ref](https://pkg.go.dev/cloud.google.com/go/pubsublite/pscompat#ReassignmentHandlerFunc)). If we track independent of consumption then we will have a lot more flexibility in handling partition based logic.

inge4pres:
> For kafka, we can use franz-go's `OnPartition{Assigned, Lost}` ([ref](https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#OnPartitionsAssigned)) methods and for pubsublite we can use the `ReassignmentHandler` ([ref](https://pkg.go.dev/cloud.google.com/go/pubsublite/pscompat#ReassignmentHandlerFunc)). If we track independent of consumption then we will have a lot more flexibility in handling partition based logic.

Trying to process your comment, I am fairly new to the codebase and the libraries so thanks fer being my rubber duck.

The goal of the PR is to allow consumers to know which partition each message is polled from. The reason to know such information is that when a client consuming from multiple partitions wants to have as many processing goroutines as partitions polled, each message can follow a different pipeline, landing in the parallel gorotuine (or processor) that is responsible to manage it.
Do you suggest that it is possible to achieve the same using the callbacks you mentioned?
This PR is specific for Kafka, but I'd be happy to extend the same concept to pubsublite.

> If we track independent of consumption then we will have a lot more flexibility in handling partition based logic.

What do you mean exactly with "track independent of consumption"?
The goal is not just tracking (e.g. with metrics) but also to allow messages' consumers to actively place each record on a dedicated processing pipeline (thread/goroutine/interface) when reading off the queue.

marclop:
@inge4pres Did you consider adding a new  `Partition` field to `apmqueue.Record`? It seems like a more explicit and robust API where the partition ID is passed down in each record. Implementations of `apmqueue. Processor` are free to do what they wish with that information during processing time.

```go
type Record struct {
	// Topics holds the topic where the record will be produced.
	Topic Topic
	// Partition holds the partition where the record is consumed from.
	Partition uint32
	// OrderingKey is an optional field that is hashed to map to a partition.
	// Records with same ordering key are routed to the same partition.
	OrderingKey []byte
	// Value holds the record's content. It must not be mutated after Produce.
	Value []byte
}
```

In PubSub Lite, the partition ID is encoded in the `Message.ID`, which can be decoded using `pscompat.ParseMessageMetadata`.

https://github.com/elastic/apm-queue/blob/18bac8091d166fd6b079f6a7906af04a0c905f06/pubsublite/consumer.go#L293

kafka/consumer.go
```diff
@@ -530,10 +534,14 @@ func (c *pc) consumeRecords(ftp kgo.FetchTopicPartition) {
 		// only the first record is received.
 		last := -1
 		for i, msg := range ftp.Records {
-			meta := make(map[string]string)
+			meta := make(map[string]string, len(msg.Headers))
```
kyungeunni, line 537: We are adding one more entry later for `metadataCtxPartitionKey` key (L543). Would it make sense to specify cap to be `len(msg.Headers) + 1`?

inge4pres: Ah thanks  👍🏼 will do

lahsivjar:
> What do you mean exactly with "track independent of consumption"?
The goal is not just tracking (e.g. with metrics) but also to allow messages' consumers to actively place each record on a dedicated processing pipeline (thread/goroutine/interface) when reading off the queue.

Ah, I misunderstood the PR a bit. I agree, the proposal in [this comment](https://github.com/elastic/apm-queue/pull/447#issuecomment-2076689327) was towards being able to track which partitions are being consumed by a consumer in advance of the consumption but it won't help towards the other objectives of the PR. Apologies about the misdirection.

inge4pres:
> Did you consider adding a new `Partition` field to `apmqueue.Record`? It seems like a more explicit and robust API where the partition ID is passed down in each record.

Good point, thanks 🙏🏼 
I had not thought about it, and followed an implementation on the same principle as `project_id` being part of the record context.
I'll try that and see how it performs👍🏼 should we also make `project_id` a part of Record then?
What if we leave both on the metadata for now and proceed in extracting them as fields of `Record` in a future moment?

inge4pres:
> should we also make `project_id` a part of Record then?

While reasoning again on this I thought: while the partition is part of the queue domain, project Id is not.
Therefore it does not make any sense to add to apm-queue the `project_id` concept, it should be treated as any other record metadata.
The partition ID, instead, is entirely part of how records are being processed (they are inherently part of a partition) so I will follow the suggestion and make if part of the `Record` struct.

kafka/consumer.go
inge4pres: I left it as-is in view of Marc's proposal, sowe don't have an additional entry in the map other than the existing headers

kafka/consumer_test.go
```diff
@@ -162,7 +163,11 @@ func TestConsumerInstrumentation(t *testing.T) {
 		MaxPollRecords: 1, // Consume a single record for this test.
 		Processor: apmqueue.ProcessorFunc(func(_ context.Context, r apmqueue.Record) error {
 			defer close(processed)
-			assert.Equal(t, event, r)
+			assert.Equal(t, event.Topic, r.Topic)
+			assert.Equal(t, event.Value, r.Value)
+			// We cannot know which partition the record will be polled from,
+			// but we can assert it's in the range of number of partitions.
+			assert.True(t, event.Partition < partitions-1)
```
marclop, line 170: Would that ever not be the case? I don't follow what's being asserted here.

kafka/consumer_test.go
```diff
@@ -593,7 +599,11 @@ func TestMultipleConsumerGroups(t *testing.T) {
 		m[gid] = &counter
 		cfg.Processor = apmqueue.ProcessorFunc(func(_ context.Context, r apmqueue.Record) error {
 			counter.Add(1)
-			assert.Equal(t, event, r)
+			assert.Equal(t, event.Topic, r.Topic)
+			assert.Equal(t, event.Value, r.Value)
+			// We cannot know which partition the record will be polled from,
+			// but we can assert it's in the range of number of partitions.
+			assert.True(t, event.Partition < partitions-1)
```
marclop, line 606: Same comment on if it's possible for this to ever be false.

marclop (commented):
Are you planning on tackling pub/sub lite on another PR? Or simply open an issue to defer it?

kafka/consumer_test.go
inge4pres: The assertion is that we're setting an ID for the field `Partition` that is in the [0,n) range.
That cannot be greater than the number of partitions - 1.

kafka/consumer_test.go
inge4pres: Yes in case the field is reset from someone modifying the Record during `Process`, since the field is exported.
We could un-export and use a getter, at that point the test can be removed.

marclop (approved):
Looks good!

inge4pres:
> Are you planning on tackling pub/sub lite on another PR? Or simply open an issue to defer it?

Will implement pubsub also in this Pr, I was waiting for a confirmation on where to place the partition Id.
We agree it goes into Record, so I'll move ahead with that.

The only downside I see is this implmentation is that we're making the struct less mechanical-sympathetic , as we are moving it from a 64-bit aligned struct to 2 words.
I don't think this poses a performance concern, but i'll keep an eye on the tests for it.

kafka/consumer_test.go
inge4pres: Oh maybe you're referring to the fact that `Partition` will always exist because of its zero-value?

marclop: I think it's good to assert there's something there and is populated, that'll always be the case. A possibility for a somewhat more thorough test would be to produce enough records that they're produced to all partitions and have a test that ensures that records are received from all the partitions.

I think that'd be a good test to add, but I'll leave it up to you if you want to do it on this PR or as a follow up.

inge4pres:
The proposal is implemented 🎉 
An additional function in `queuecontext`, with signature "Enrich(context.Context, string, string)" is added as a utility to transmute the field from Record into context.

marclop (approved):
LGTM

PR #455, opened 2024-05-06
feat: add metric for compressed/uncompressed bytes sent to kafka

kruskall:
Add metrics to track compressed and uncompressed bytes sent to kafka.

- `producer.messages.bytes`
- `producer.messages.uncompressed_bytes`


vikmik (approved):
:+1: perfect, thanks!

PR #460, opened 2024-05-08
feat: rework kotel metrics

kruskall:
Remove kotel plugin and publish those metrics in our own hook.
Add compressed/uncompressed fetched bytes.
Use consistent attributes across metrics.

This PR keeps the same name for the kotel metrics for backward comaptibility.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	cloud.google.com/go/monitoring v1.19.0
 	cloud.google.com/go/pubsub v1.38.0
 	cloud.google.com/go/pubsublite v1.8.1
-	github.com/aws/aws-sdk-go-v2/config v1.27.12
+	github.com/aws/aws-sdk-go-v2/config v1.27.13
 	github.com/google/go-cmp v0.6.0
 	github.com/googleapis/gax-go/v2 v2.12.4
 	github.com/stretchr/testify v1.9.0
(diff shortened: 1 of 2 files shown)


kafka/metrics.go
```diff
@@ -246,3 +528,10 @@ func attributesFromRecord(r *kgo.Record, extra ...attribute.KeyValue) []attribut
 	}
 	return attrs
 }
+
+func strnode(node int32) string {
```
endorama, line 532: question: could you clarify why this is needed and why the return value (seems to) differs between `node < 0` and `node >= 0`?

kafka/metrics.go
```diff
@@ -60,11 +70,31 @@ type metricHooks struct {
...
 
-	messageProduced          metric.Int64Counter
-	messageFetched           metric.Int64Counter
-	messageDelay             metric.Float64Histogram
-	messageBytes             metric.Int64Counter
-	messageUncompressedBytes metric.Int64Counter
+	// kotel metrics
+
+	connects    metric.Int64Counter
+	connectErrs metric.Int64Counter
```
endorama, line 76: question: as we submitted a PR to update these metrics (https://github.com/twmb/franz-go/pull/691) do you think it make sense to keep compatibility with `kotel` here? I don't see benefits in the long term but I'm ok if we want to keep compatibility for this PR

kafka/metrics_test.go
```diff
@@ -386,11 +455,24 @@ func TestConsumerMetrics(t *testing.T) {
 		},
 	}
 
+	ignore := []string{
```
kyungeunni, line 458: Why do we need to ignore them? should we be able to check these metrics too as it's now reported by our own hook?

kyungeunni (commented):
Looks great, may I ask what motivated the removal of kotel? Is there a ticket I can refer to have more context?

kafka/metrics.go
```diff
@@ -73,43 +101,168 @@ func newKgoHooks(mp metric.MeterProvider, namespace, topicPrefix string,
 	topicAttributeFunc TopicAttributeFunc,
 ) (*metricHooks, error) {
 	m := mp.Meter(instrumentName)
+
+	// kotel metrics
+
+	// connects and disconnects
+	connects, err := m.Int64Counter(
+		"messaging.kafka.connects.count",
```
kyungeunni, line 109: Q. Should we define constant values for the kotel metrics, too? On the other hands, it also make sense to inline them to distinguish from our custom metric.

kafka/metrics.go
twmb: FWIW I plan to merge 691 actually sooner than later, if that's the problem.

kafka/metrics.go
kyungeunni: Makes sense. thanks for the clarification 👍

kyungeunni (approved):

PR #468, opened 2024-05-13
fix: add missing namespace and topic attributes

kruskall:
Add missing namespace to kotel metrics
Add missing topic attributes to our own metrics (not sure why these were missing)

### DIFF
--- a/kafka/manager_test.go
+++ b/kafka/manager_test.go
@@ -156,12 +156,15 @@ func TestManagerDeleteTopics(t *testing.T) {
 	assert.Empty(t, cmp.Diff(metrictest.Int64Metrics{
 		{Name: "messaging.kafka.connects.count", Unit: "1"}: {
 			{K: "messaging.system", V: "kafka"}: 2,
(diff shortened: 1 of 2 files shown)


kafka/metrics.go
```diff
@@ -311,70 +311,78 @@ func formatMetricError(name string, err error) error {
...
 
 func (h *metricHooks) OnBrokerConnect(meta kgo.BrokerMetadata, _ time.Duration, _ net.Conn, err error) {
-	attributes := attribute.NewSet(
-		semconv.MessagingSystem("kafka"),
-	)
+	attrs := make([]attribute.KeyValue, 0, 2)
+	attrs = append(attrs, semconv.MessagingSystem("kafka"))
+	if h.namespace != "" {
+		attrs = append(attrs, attribute.String("namespace", h.namespace))
+	}
```
vikmik, line 318: It seems like this could be extracted into a function to avoid the 4x repetition. Up to you.

vikmik (approved):

Issue #464, opened 2024-05-10
Tests: flaky failures


inge4pres:
**Describe the current state**

During the development of #447 I noticed some flaky tests, passing locally most of the times but not in CI:

* consumer_test.go:224 -> ProducedWithHeaders
```
    metrics_test.go:64: [Metrics Data not equal: Sum[int64] not equal: Sum DataPoints not equal:
        missing expected values:
```

* consumer_test.go:464 --- FAIL: TestConsumerGracefulShutdown (0.03s)
```
    --- FAIL: TestConsumerGracefulShutdown/AtLeastOnceDelivery (0.01s)
    --- PASS: TestConsumerGracefulShutdown/AtMostOnceDelivery (0.02s)
    
            	Error Trace:	/home/runner/work/apm-queue/apm-queue/kafka/consumer_test.go:464
        	            				/home/runner/work/apm-queue/apm-queue/kafka/consumer_test.go:469
        	Error:      	Not equal: 
        	            	expected: 2
        	            	actual  : 1
        	Test:       	TestConsumerGracefulShutdown/AtLeastOnceDelivery
```

* TestProducerMetrics

**Describe the desired state**

The tests are consistently reporting either a failure or a success.

inge4pres:
For `TestProducerMetrics`: running it with 

```
 go test -v -count=10 -run="TestProducerMetrics/DeadlineExceeded" ./kafka/...
```

most certainly produces a panic

```
=== RUN   TestProducerMetrics/DeadlineExceeded
    metrics_test.go:65: 
                Error Trace:    /home/francesco/Projects/github.com/elastic/apm-queue/kafka/metrics_test.go:65
                                                        /home/francesco/Projects/github.com/elastic/apm-queue/kafka/metrics_test.go:123
                Error:          Not equal: 
                                expected: 2
                                actual  : 1
                Test:           TestProducerMetrics/DeadlineExceeded
    metrics_test.go:67: producer.messages.count
    metrics_test.go:71: [Name not equal:
        expected: messaging.kafka.connect_errors.count
        actual: producer.messages.count Description not equal:
        expected: Total number of connection errors, by broker
        actual: The number of messages produced Metrics Data not equal: Sum[int64] not equal: Sum DataPoints not equal:
        missing expected values:
[...]

PR #469, opened 2024-05-13
test: remove connection count assertions

kruskall:
We can't assert the connection count as it's not deterministic

I suspect the metadata request run during client creation might also screw the metrics, causing flakiness.


inge4pres (approved):
Awesome I think this fixes partially #464

PR #479, opened 2024-05-23
Align connects metrics

endorama:
This PR address https://github.com/twmb/franz-go/issues/670 but in the monitoring implementation introduced in this module by #460.

It changes Kafka connection metrics, instead of relying on 2 different metrics to report Kafka connections successes and failures, use only a single metric with distinct `outcome` attribute based on the outcome. This implementation does not guard this new behaviour under a flag, like I did for the contribution to `franz-go` as I don't think it's useful is a specialized implementation like we have here. But still this introduces a breaking change in the `kafka` package monitoring so I'm open to revisit this decision if we want to keep backwards compatibility.

Metric `messaging.kafka.connect_errors.count` has been removed.

Attribute `outcome` has been added to `messaging.kafka.connects.count` with 2 possible values: `success` and `failure`.

### DIFF
--- a/kafka/manager_test.go
+++ b/kafka/manager_test.go
@@ -157,6 +157,7 @@ func TestManagerDeleteTopics(t *testing.T) {
 		{Name: "messaging.kafka.connects.count", Unit: "1"}: {
 			{K: "messaging.system", V: "kafka"}: 2,
 			{K: "namespace", V: "name_space"}:   2,
+			{K: "outcome", V: "success"}:        2,
 		},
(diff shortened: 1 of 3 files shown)


endorama:
Upon internal feedback this PR now does not remove a metric but only adds dimensions & emit `connects` metric also on failure. Removing the old metric will happen in a future update.

endorama:
Added a follow up issue to keep track of removing `messaging.kafka.connect_errors.count`.

PR #484, opened 2024-05-27
feat: add messaging.kafka.write.latency metric

### LINKED ISSUE #252: producer(kafka): Write custom hook to record batch write duration
## Description

Currently, APM queue isn't recording the time it takes to write a batch of records to the queue. It would be best to also record the time it takes to write a batch to Kafka, so clients can use latency metrics.

We can implement the interface below and record the `writeWait + timeToWrite` as `producer.write.latency` and return immediately if `key != kmsg.Produce`.

```go 
// HookBrokerWrite is called after a write to a broker.
//
// Kerberos SASL does not cause write hooks, since it directly writes to the
// connection.
type HookBrokerWrite interface {
	// OnBrokerWrite is passed the broker metadata, the key for the request
	// that was written, the number of bytes that were written (may not be
	// the whole request if there was an error), how long the request
	// waited before being written (including throttling waiting), how long
	// it took to write the request, and any error.
	//
	// The bytes written does not count any tls overhead.
	OnBrokerWrite(meta BrokerMetadata, key int16, bytesWritten int, writeWait, timeToWrite time.Duration, err error)
}
```

## Links

- https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#HookBrokerWrite

### LINKED ISSUE #483: test: broken `TestProducerMetrics`
While working on #478, I noticed that the test case `TestProducerMetrics` isn't validating anything, even specifying non-existing metric name as an expected metric, it passes.

For example, 
This test case passes:
https://github.com/elastic/apm-queue/blob/89aef429e73b767c9c58d3e3537d27c2eeefaafc/kafka/metrics_test.go#L207-L209

even though the expected unit is "Byte":
https://github.com/elastic/apm-queue/blob/89aef429e73b767c9c58d3e3537d27c2eeefaafc/kafka/metrics.go#L224-L226

Tried to fix this while working the PR, but fix to the tests cascaded other tests to fail so it seems like it requires dedicated time to fix.

kyungeunni:
closes #252 
partially fixes #483 

- Add `messaging.kafka.write.latency` metric using `OnBrokerWrite` hook with dimension of `op` (stringified request key). This will allow us to observe more closely if/when Kafka write operations take a long time or if they fail as well, as well as how long it took to do _produce_ operation.
- partially fixes `TestProducerMetrics`. It correctly asserts `want` metrics.


kafka/metrics.go
```diff
@@ -40,6 +41,7 @@ const (
 	msgProducedCountKey             = "producer.messages.count"
 	msgProducedBytesKey             = "producer.messages.bytes"
 	msgProducedUncompressedBytesKey = "producer.messages.uncompressed.bytes"
+	messageWriteLatencyKey          = "producer.messages.write.latency"
```
marclop, line 44: This metric name doesn't seem right. `OnBrokerWrite` is issued per broker, and data is written to the Kafka broker at any time.

- https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#HookBrokerWrite

For this reason, we should probably change the metric name to something like `messaging.kafka.write.latency`, since it may or may not be producing records for the request.

`OnBrokerWrite` does receive the request key metadata, so we could use the string name for the key ID and add it as a dimension to `OnBrokerWrite` metrics, see:

- https://pkg.go.dev/github.com/twmb/franz-go/pkg/kmsg#Produce

kafka/metrics.go
```diff
@@ -382,6 +394,15 @@ func (h *metricHooks) OnBrokerWrite(meta kgo.BrokerMetadata, _ int16, bytesWritt
 		int64(bytesWritten),
 		metric.WithAttributeSet(attribute.NewSet(attrs...)),
 	)
+
+	// Record latency when the request type is Produce.
+	if key == int16(kmsg.Produce) {
```
marclop, line 399: This won't record the latency for failed writes, shouldn't those be recorded too?

kafka/metrics.go
kyungeunni: `if key == int16(kmsg.Produce) ` condition was added in L399 so that we only record `produce` write latency. What I understood from the original issue was to filter out all the other requests sent to broker but record only produce latency. But if we are interested in broker latency in general, happy to make the change.

kafka/metrics.go
kyungeunni: >This won't record the latency for failed writes

Could you elaborate a bit more? I don't think the key changes depending on the result of the write, but I may have missed something.

kafka/metrics.go
marclop: Ah, that is the case; I had forgotten about the issue's contents; thanks for reminding me. It seems to have been the original intent, but considering it a bit more, it seems more future-proof and feature-complete to record latency using the stringified operation key as the dimension. This allows us to visualize more closely if/when Kafka write operations take a long time or if they fail as well, not only the produce operation.

It would also be beneficial to then open a follow-up issue to also cover https://pkg.go.dev/github.com/twmb/franz-go/pkg/kgo#HookBrokerRead.

What do you think?

kafka/metrics.go
marclop: If `err != nil` (on line 372) the code takes another branch above and returns immediately.

kafka/metrics.go
kyungeunni: I see, that sounds good 👍 I will update the PR to address this.

kafka/metrics.go
kyungeunni: Ah, I missed that, thanks!

marclop (commented):
Looks pretty good! Just a couple of minor comments

kafka/metrics.go
```diff
@@ -363,12 +375,21 @@ func (h *metricHooks) OnBrokerDisconnect(meta kgo.BrokerMetadata, _ net.Conn) {
...
+func (h *metricHooks) OnBrokerWrite(meta kgo.BrokerMetadata, key int16, bytesWritten int, writeWait, timeToWrite time.Duration, err error) {
+	attrs := make([]attribute.KeyValue, 0, 3)
+	attrs = append(attrs,
+		semconv.MessagingSystem("kafka"),
+		attribute.String("op", kmsg.NameForKey(key)),
+	)
 	if h.namespace != "" {
 		attrs = append(attrs, attribute.String("namespace", h.namespace))
 	}
+	h.messageWriteLatency.Record(
```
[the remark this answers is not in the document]
kyungeunni: Added outcome dimension: [063b8b1](https://github.com/elastic/apm-queue/pull/484/commits/063b8b159e94dd75e166c5a9a116793988df54c3)

kafka/metrics.go
```diff
@@ -363,12 +375,21 @@ func (h *metricHooks) OnBrokerDisconnect(meta kgo.BrokerMetadata, _ net.Conn) {
...
 }
 
-func (h *metricHooks) OnBrokerWrite(meta kgo.BrokerMetadata, _ int16, bytesWritten int, _, _ time.Duration, err error) {
-	attrs := make([]attribute.KeyValue, 0, 2)
-	attrs = append(attrs, semconv.MessagingSystem("kafka"))
+func (h *metricHooks) OnBrokerWrite(meta kgo.BrokerMetadata, key int16, bytesWritten int, writeWait, timeToWrite time.Duration, err error) {
+	attrs := make([]attribute.KeyValue, 0, 3)
+	attrs = append(attrs,
+		semconv.MessagingSystem("kafka"),
+		attribute.String("op", kmsg.NameForKey(key)),
```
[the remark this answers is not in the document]
kyungeunni: addressed [063b8b1](https://github.com/elastic/apm-queue/pull/484/commits/063b8b159e94dd75e166c5a9a116793988df54c3#diff-46f9482b8dc116dd9295c4749514961aff40d3528a5706aaa240dafbf31a0b4dR382)

marclop (approved):
Looking great! Thanks for the changes.

PR #537, opened 2024-08-06
test(kafka): Fix unit test flakiness

marclop:
- `TestNewProducerBasic`: Remove the need to use a mutex in the test by closing the producer before performing the assertions.
- `TestConsumerGracefulShutdown`: Improve test synchronization.
- `TestManagerDeleteTopics`: Only assert deletion metrics, not necessarily written.

### DIFF
--- a/kafka/consumer_test.go
+++ b/kafka/consumer_test.go
@@ -445,6 +445,7 @@ func TestConsumerGracefulShutdown(t *testing.T) {
 		var processed atomic.Int32
 		var errored atomic.Int32
 		process := make(chan struct{})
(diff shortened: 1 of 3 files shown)


marclop:
A few PRs seem to be failing due to these flaky tests. There may be more, but these are the biggest offenders.

simitt:
@rubvs one of the tests that @marclop fixed here will also fix https://github.com/elastic/apm-queue/issues/464. Given that you had started looking into this issue already, I suggest you will give this PR a review and comment if you had found anything additional in the test or alternative suggestions for fixes.

Issue #540, opened 2024-08-09
Allow enabling `franz-go` logging from `kafka` package


endorama:
The `kafka` package does not initialize `franz-go` logging nor exposes a way to do this. This means there is no way for callers to enable `franz-go` logging.

## Goal

Allow enabling `franz-go` logging from `kafka` package callers.

inge4pres:
Clarification: when we decide to pursue logging the _entirety_ of messages from franz-go, we can remove what's been added in #541.
If we want to make both approach co-exists, it would be sensible to somehow configure franz-go logging, meaning:

 - log level
 - filters to include/exclude entries at ingestion time

The previous note is assuming that we want to reduce the amount of data stored in our logging infrastructure, or at least limit to what's actually useful.

raultorrecilla:
moving this from it105 to 109

PR #541, opened 2024-08-09
Kafka: add hook to log connection error reasons

inge4pres:
### Reason for this PR

The current `metricsHook` struct that collects metrics based on Kafka hooks is not equipped to also report details of _why_ connection errors are caused.

### Details

We provide a new struct that implements a (single, for now) kgo Hook dedicated to logging errors during the broker connection phase.
The struct can be iterated in the future to implement other monitoring hooks, providing logging utilities where the metrics would have too high cardinality.

### Additional context

Currently, the proposed design adds the new logger hook behind the `CommonConfig.DisableTelemetry` flag (similarly to the metrics hook).
This can be improved and it is not a perfect solution, but it has to be done this way until a bigger refactor is done to expose the `CommonConfig.hook` fieldin some way.

The more idiomatic, but also more expensive, solution would be to allow exporting franz-go logs into the Logger.
This is mentioned in https://github.com/elastic/apm-queue/issues/540

### DIFF
--- a/kafka/logger.go
+++ b/kafka/logger.go
@@ -1,3 +1,20 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
 package kafka
 
 import (
--- a/kafka/logger.go
+++ b/kafka/logger.go
@@ -16,11 +16,13 @@ type loggerHook struct {
 }
 
 // OnBrokerConnect implements the kgo.HookBrokerConnect interface.
-func (l *loggerHook) OnBrokerConnect(_ kgo.BrokerMetadata, dialDur time.Duration, _ net.Conn, err error) {
+func (l *loggerHook) OnBrokerConnect(meta kgo.BrokerMetadata, dialDur time.Duration, _ net.Conn, err error) {
 	if err != nil {
 		l.logger.Error("failed to connect to broker",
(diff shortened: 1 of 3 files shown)


kafka/logger.go
```diff
@@ -0,0 +1,45 @@
...
+
+type loggerHook struct {
+	logger *zap.Logger
+}
+
+// OnBrokerConnect implements the kgo.HookBrokerConnect interface.
+func (l *loggerHook) OnBrokerConnect(meta kgo.BrokerMetadata, dialDur time.Duration, _ net.Conn, err error) {
+	if err != nil {
+		l.logger.Error("failed to connect to broker",
+			zap.Error(err),
```
1pkg, line 39: Do you think it's also worthwhile to add `zap.Stack` to this log entry? There are multiple code paths that could call into connect/reconnect in franz-go, it would be nice to record them.

1pkg (dismissed):

kafka/logger_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"go.uber.org/zap/zaptest/observer"
+)
+
+func TestHookLogsFailedDial(t *testing.T) {
+	cluster, cfg := newFakeCluster(t)
+	t.Cleanup(cluster.Close)
+
+	core, logs := observer.New(zap.ErrorLevel)
+	cfg.Logger = zap.New(core)
+	//cfg.hooks = []kgo.Hook{&loggerHook{logger: cfg.Logger}}
```
endorama, line 39:
```suggestion
```

inge4pres: Thanks, slipped it

kafka/logger.go
inge4pres: Makes sense, thanks 👍🏼 
Hopefully the stacktrace is not generating too much output, since this is an Error level, it will be present in all failed connect attempts, and AFAIU from investigation made by @endorama we see a lot of them.
I'll add it, and we can remove later on if we see it's too verbose.

inge4pres:
Unpleasant discovery: the newly added test is currently flaky because of franz-go implcitly retrying to reconnect at least once, failure here.
https://github.com/elastic/apm-queue/actions/runs/10324596296/job/28584375960?pr=541#step:4:8413

I think we should disable the retry to make the test consistent but I am unsure how to pass that option to the client 🤔

inge4pres:
> I think we should disable the retry to make the test consistent but I am unsure how to pass that option to the client 🤔

Not seeing an option to control the connection retry, I ended up changing the assertion: we now validate that there is either a single or 2 log lines containing the error

1pkg (approved):

endorama (approved):

Issue #562, opened 2024-09-04
Flaky pubsublite test


rubvs:
While investigation https://github.com/elastic/apm-queue/issues/464 I came across a more subtle flaky test in the `pubsublite` package:

```
apm-queue [main] > make test
go test -race  -count=10 -timeout=60s ./...
?   	github.com/elastic/apm-queue/v2	[no test files]
?   	github.com/elastic/apm-queue/v2/cmd/queuebench	[no test files]
?   	github.com/elastic/apm-queue/v2/metrictest	[no test files]
ok  	github.com/elastic/apm-queue/v2/kafka	18.965s
ok  	github.com/elastic/apm-queue/v2/pubsublite	3.949s
ok  	github.com/elastic/apm-queue/v2/pubsublite/internal/pubsubabs	3.205s
--- FAIL: TestPublisher (0.02s)
    --- FAIL: TestPublisher/success_with_otel_attributes#01 (0.01s)
        publisher_test.go:266:
            	Error Trace:	/Users/rubenvanstaden/github/elastic/apm-queue/pubsublite/internal/telemetry/publisher_test.go:266
            	Error:      	Not equal:
[...]

PR #563, opened 2024-09-04
Fix flaky pubsublite test

rubvs:
Fixes https://github.com/elastic/apm-queue/issues/562

This is not a proper fix, but I think it's a hacky improvement.

The problem with the hack in the unit test in https://github.com/elastic/apm-queue/blob/6675642e184c322eadac10a3c6155ed267fc8bdb/pubsublite/internal/telemetry/publisher_test.go#L242 is that the "logic" is tied to the unit test, not in the actual implementation. 

Though, this PR does not fully fix the issue, I believe that this PR is a more idiomatic implementation, until we can find a proper solution.

At the core the issue with the goroutine in https://github.com/elastic/apm-queue/blob/6675642e184c322eadac10a3c6155ed267fc8bdb/pubsublite/internal/telemetry/publisher.go#L114 is that we fire-and-forget about it.

The proper solution is to either:

1. Use wait-group, which to my understanding defeats the purpose of a fire-and-forget strategy
2. Orchestrate communication between the child and parent, which means we have to update the function interface, by either passing or returning a channel.

### DIFF
--- a/pubsublite/internal/pubsubabs/publisher.go
+++ b/pubsublite/internal/pubsubabs/publisher.go
@@ -21,6 +21,7 @@ package pubsubabs
 
 import (
 	"context"
+	"time"
 
 	"cloud.google.com/go/pubsub"
 )
@@ -129,4 +130,6 @@ func SetPublishResult(r *IPublishResult, sid string, err error) {
 	r.serverID = sid
 	r.err = err
 	close(r.ready)
+	// Wait for fire-and-forget goroutine in producer.Publish
+	<-time.After(5 * time.Millisecond)
 }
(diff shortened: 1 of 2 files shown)


axw:
I think we could probably consider deleting the pubsublite implementation, since Pub/Sub Lite has been deprecated (see https://cloud.google.com/pubsub/docs/choosing-pubsub-or-lite) and [Apache Kafka for BigQuery](https://cloud.google.com/products/managed-service-for-apache-kafka) has been announced. WDYT @marclop?

marclop:
I'll open a PR to remove pubsublite

marclop:
@rubvs I think we can close this one. GCP PubSub Lite is no more 🙂

PR #579, opened 2024-10-03
metrics: add consumer read latency distribution metric

1pkg:
# Description

This PR adds missing latency histogram metric for consumer `HookBrokerRead` hook in `franz-go`.
This PR additionally fixes broken `TestConsumerMetrics` which was always skipping test assertions.


kafka/metrics_test.go
```diff
@@ -505,8 +524,17 @@ func TestConsumerMetrics(t *testing.T) {
 	metrics := filterMetrics(t, rm.ScopeMetrics)
 	assert.Len(t, metrics, len(wantMetrics)+len(ignore))
 
-	for k, m := range wantMetrics {
-		metric := metrics[k]
```
1pkg, line 509: This was a faulty test, the actual produced metrics should be matched against expected metrics using the metric name.

kafka/metrics.go
```diff
@@ -407,23 +419,31 @@ func (h *metricHooks) OnBrokerWrite(meta kgo.BrokerMetadata, key int16, bytesWri
 	)
 }
 
-func (h *metricHooks) OnBrokerRead(meta kgo.BrokerMetadata, _ int16, bytesRead int, _, _ time.Duration, err error) {
+func (h *metricHooks) OnBrokerRead(meta kgo.BrokerMetadata, _ int16, bytesRead int, readWait, timeToRead time.Duration, err error) {
 	attrs := make([]attribute.KeyValue, 0, 2)
```
marclop, line 423: To avoid an extra allocation, I think we should bump the slice capacity from 2 to 3
```suggestion
	attrs := make([]attribute.KeyValue, 0, 3)
```

marclop (approved):

kafka/metrics.go
1pkg: updated

kafka/metrics_test.go
```diff
@@ -65,22 +63,16 @@ func TestProducerMetrics(t *testing.T) {
 		assert.NoError(t, rdr.Collect(context.Background(), &rm))
 
 		metrics := filterMetrics(t, rm.ScopeMetrics)
-		for i := range metrics {
-			t.Log(metrics[i].Name)
-		}
-		for i := range want {
-			if slices.Contains(ignore, want[i].Name) {
```
1pkg, line 72: I also refactored `ignored` metrics bit of the test here so it should fully address the concerns from https://github.com/elastic/apm-queue/issues/483

marclop (dismissed):
Thanks!

kafka/metrics.go
```diff
@@ -271,6 +274,14 @@ func newKgoHooks(mp metric.MeterProvider, namespace, topicPrefix string,
 		return nil, formatMetricError(msgConsumedUncompressedBytesKey, err)
 	}
 
+	messageReadLatency, err := m.Float64Histogram(messageReadLatencyKey,
+		metric.WithDescription("Time took to read a batch including waited before being read"),
```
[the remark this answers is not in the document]
1pkg: I used the same confusing wording from `messaging.kafka.write.latency`. Now updated both for consistency

PR #593, opened 2024-11-06
`producer.messages.count`: Use `kgo.Err` as reason

marclop:
Updates the `error_reason` dimension to use the value from the returned error (`kgo.Err.Message`). This allows us to classify the errors with the reason more granularly.

Previously, all the produce failures that aren't due to `context.Done()` would be classified as `unknown`.


carsonip (dismissed):

carsonip:
Can you also update the docs on various services that depend on apm-queue?

carsonip (approved):

PR #618, opened 2025-01-16
fix: better protect against nil topicAttributeFunc

kruskall:
## ❓ Why is this being changed

systemtests are failing because of a nil topic attribute func. The topic manager does not handle nil func.

See https://github.com/elastic/apm-queue/actions/runs/12741942829/job/35509307227
 
## 🧑‍💻 What is being changed

set the default func in the common config instead of the metrics hook
 
## ✅ How to validate the change

run systemtests


rubvs (approved):

marclop:
@kruskal I also think it makes sense to not open PRs from your fork, since that skips the `systemtests`

PR #625, opened 2025-01-28
TLS: Add support for CA certificate in Kafka

marclop:
This commit adds support for setting the TLS CA in the Kafka TLS configuration.

It supports hot reloading the certificate if it's updated.


kafka/common_test.go
```diff
@@ -318,3 +324,66 @@ func TestTopicFieldFunc(t *testing.T) {
...
+		KeyUsage:     x509.KeyUsageCertSign | x509.KeyUsageKeyEncipherment,
+		IsCA:         true,
+	}
+	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, &key.PublicKey, key)
+	require.NoError(t, err)
+
+	return pemEncodeCert(derBytes)
+}
+
+func pemEncodeCert(certDER []byte) []byte {
```
1pkg, line 347: nit: slightly cleaner way to achieve the same is using [pem.EncodeToMemory](https://pkg.go.dev/encoding/pem#EncodeToMemory).

1pkg (approved):

kafka/common_test.go
```diff
@@ -318,3 +324,59 @@ func TestTopicFieldFunc(t *testing.T) {
...
+		KeyUsage:     x509.KeyUsageCertSign | x509.KeyUsageKeyEncipherment,
+		IsCA:         true,
+	}
+	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, &key.PublicKey, key)
+	require.NoError(t, err)
+
+	return pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
+}
+
+func TestTLSCACertPath(t *testing.T) {
```
axw, line 347: I think it would be good to add a test that verifies the CA is effective when dialling too. The code is simple enough that we could leave that till later though.

kafka/common.go
```diff
@@ -215,6 +216,21 @@ func (cfg *CommonConfig) finalize() error {
 		if os.Getenv("KAFKA_TLS_INSECURE") == "true" {
 			cfg.TLS.InsecureSkipVerify = true
 		}
+		if caCertPath := os.Getenv("KAFKA_TLS_CA_CERT_PATH"); caCertPath != "" {
+			caCert, err := os.ReadFile(caCertPath)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("kafka: failed to read CA cert: %w", err))
+			} else {
+				rootCa, err := x509.SystemCertPool()
```
axw, line 224: I wonder if we should _only_ accept the configured CA? Then clients would have more control

kafka/common.go
```diff
@@ -215,6 +216,21 @@ func (cfg *CommonConfig) finalize() error {
 		if os.Getenv("KAFKA_TLS_INSECURE") == "true" {
 			cfg.TLS.InsecureSkipVerify = true
 		}
+		if caCertPath := os.Getenv("KAFKA_TLS_CA_CERT_PATH"); caCertPath != "" {
+			caCert, err := os.ReadFile(caCertPath)
```
axw, line 220: Do we need to support reloading the CA cert file, in case of rotation?

axw (commented):
Looks good, just a few questions.

kafka/common.go
marclop: true, I think in this case, we may need to close the client and trigger a pod restart, since I don't think there are the mechanisms in place for hot reloading atm.

kafka/common.go
axw: You would not be at the mercy of root CAs, which can be compromised. Rare I know, but it happens, e.g. https://blog.mozilla.org/security/2011/03/25/comodo-certificate-issue-follow-up/

kafka/common.go
axw: One option I think would be to create a Dialer that dynamically loads the CA cert each time it dials. I haven't thought through that deeply though

kafka/common_test.go
```diff
@@ -318,3 +328,160 @@ func TestTopicFieldFunc(t *testing.T) {
...
+
+		t.Setenv("KAFKA_TLS_CA_CERT_PATH", tempFile)
+		cfg := CommonConfig{Brokers: []string{"broker"}, Logger: zap.NewNop()}
+		err = cfg.finalize()
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "kafka: error creating dialer with CA cert")
+	})
+}
+
+func TestTLSHotReload(t *testing.T) {
```
axw, line 392: Just a thought: maybe this test could be simplified by using httptest? httptest.NewTLSServer creates its own certificate, and you can extract the CA out of the `Server.Client()`'s TLS config.

kafka/common_test.go
```diff
@@ -318,3 +328,160 @@ func TestTopicFieldFunc(t *testing.T) {
...
+	tempFile := filepath.Join(t.TempDir(), fmt.Sprintf("%s.pem", rand.Text()))
```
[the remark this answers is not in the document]
marclop: When running the test with `-count`, it seemed to fail when the file was the same, so I assumed some sort of contention on that file name. Changing it to a random name made it better.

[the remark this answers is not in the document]
axw: `t.TempDir()` returns a new temp dir on each call, so that doesn't make any sense to me.

What failures were you observing? I just tried running the test locally and it fails even without `-count` due to the `time.After` being outside of the accept loop. I have 8 cores, 16 threads, so that loop will take a little while, and will be non-deterministic. If I set GOMAXPROCS to something lower then it passes.

kafka/common.go
```diff
@@ -355,3 +368,68 @@ func topicFieldFunc(f TopicLogFieldFunc) TopicLogFieldFunc {
...
 }
+
+// newCertReloadingDialer returns a dialer that reloads the CA cert when the
+// file mod time changes.
+func newCertReloadingDialer(caPath string, tlsCfg *tls.Config) (func(ctx context.Context, network, address string) (net.Conn, error), error) {
+	p, err := os.Stat(caPath)
+	if err != nil {
+		return nil, err
+	}
+	dialer := &net.Dialer{Timeout: 10 * time.Second}
```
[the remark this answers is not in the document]
marclop: That's right, added the comment.

kafka/common.go
```diff
@@ -215,6 +218,16 @@ func (cfg *CommonConfig) finalize() error {
 		if os.Getenv("KAFKA_TLS_INSECURE") == "true" {
 			cfg.TLS.InsecureSkipVerify = true
 		}
+		if caCertPath := os.Getenv("KAFKA_TLS_CA_CERT_PATH"); caCertPath != "" {
```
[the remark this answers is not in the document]
marclop: Added some checks and unit tests.

kafka/common_test.go
marclop: Refactored and re-used the httptest.server TLS certificates now.

axw (approved):

PR #636, opened 2025-02-25
Kafka: Support specifying client certificates

marclop:
Adds support for Kafka to use client certificates based on two new env variables `KAFKA_TLS_CERT_PATH` and `KAFKA_TLS_KEY_PATH`. The current implementation supports hot reloading by checking if the file write time is greater than the last time it was read.

Most of the changes are in the form of unit tests.

### DIFF
--- a/kafka/common.go
+++ b/kafka/common.go
@@ -220,7 +220,7 @@ func (cfg *CommonConfig) finalize() error {
 		caCertPath := os.Getenv("KAFKA_TLS_CA_CERT_PATH")
 		if tlsInsecure && (caCertPath != "" || certPath != "" || keyPath != "") {
 			errs = append(errs, errors.New(
-				"kafka: cannot set both KAFKA_TLS_INSECURE and KAFKA_TLS_CA_CERT_PATH, KAFKA_TLS_CERT_PATH, or KAFKA_TLS_KEY_PATH",
+				"kafka: cannot set KAFKA_TLS_INSECURE when either of KAFKA_TLS_CA_CERT_PATH, KAFKA_TLS_CERT_PATH, or KAFKA_TLS_KEY_PATH are set",
 			))
 			break
 		}
--- a/kafka/common.go
+++ b/kafka/common.go
@@ -27,7 +27,6 @@ import (
 	"os"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


kafka/common.go
```diff
@@ -217,19 +218,20 @@ func (cfg *CommonConfig) finalize() error {
 		cfg.TLS = &tls.Config{}
 		tlsInsecure := os.Getenv("KAFKA_TLS_INSECURE") == "true"
 		caCertPath := os.Getenv("KAFKA_TLS_CA_CERT_PATH")
-		if tlsInsecure && caCertPath != "" {
+		if tlsInsecure && (caCertPath != "" || certPath != "" || keyPath != "") {
 			errs = append(errs, errors.New(
-				"kafka: cannot set both KAFKA_TLS_INSECURE and KAFKA_TLS_CA_CERT_PATH",
+				"kafka: cannot set both KAFKA_TLS_INSECURE and KAFKA_TLS_CA_CERT_PATH, KAFKA_TLS_CERT_PATH, or KAFKA_TLS_KEY_PATH",
```
inge4pres, line 223: the message is a little hard to read, do I understand correctly that INSECURE is mutually exclusive with any of the other 3?
If that's the case, what do you think of

```suggestion
				"kafka: cannot set KAFKA_TLS_INSECURE when either of KAFKA_TLS_CA_CERT_PATH, KAFKA_TLS_CERT_PATH, or KAFKA_TLS_KEY_PATH are set",
```

kafka/common.go
```diff
@@ -217,19 +218,20 @@ func (cfg *CommonConfig) finalize() error {
...
 		if tlsInsecure {
 			cfg.TLS.InsecureSkipVerify = true
 		}
 		if caCertPath != "" {
-			// Auto-configure a dialer that reloads the CA cert when the file
-			// changes.
-			dialFn, err := newCertReloadingDialer(caCertPath, cfg.TLS)
+			// Set a dialer that reloads the CA cert when the file changes.
+			dialFn, err := newCertReloadingDialer(
+				caCertPath, certPath, keyPath, 30*time.Second, cfg.TLS,
```
inge4pres, line 233: Question: what happens in between the time that the certificate changed and it is reloaded?
Do we risk to lose data because the connection is rejected?

inge4pres (approved):

Issue #661, opened 2025-05-16
Updating kfake fails TestConsumerDelivery


rubvs:
A new error is observed when updating franz-go and kfake, see [Failing PR](https://github.com/elastic/apm-queue/pull/621).

```
logger.go:146: 2025-05-28T10:39:54.773-0400     ERROR   1.kafka.partition       data loss: unable to process event      {"namespace": "name_space", "topic": "topic", "partition": 0, "error": "failed processing record", "offset": 0, "headers": {}}
logger.go:146: 2025-05-28T10:39:54.774-0400     ERROR   1.kafka.partition       data loss: unable to process event      {"namespace": "name_space", "topic": "topic", "partition": 0, "error": "failed processing record", "offset": 1, "headers": {}}
logger.go:146: 2025-05-28T10:39:54.774-0400     ERROR   1.kafka.partition       data loss: unable to process event      {"namespace": "name_space", "topic": "topic", "partition": 0, "error": "extra records are lost", "offset": 2, "headers": {}}
logger.go:146: 2025-05-28T10:39:54.775-0400     ERROR   1.kafka.partition       data loss: unable to process event      {"namespace": "name_space", "topic": "topic", "partition": 0, "error": "extra records are lost", "offset": 3, "headers": {}}
logger.go:146: 2025-05-28T10:39:54.775-0400     INFO    1.kafka.kafka   assigning partitions    {"namespace": "name_space", "why": "invalidating all assignments in LeaveGroup", "how": "unassigning everything", "input": ""}
logger.go:146: 2025-05-28T10:39:54.775-0400     INFO    1.kafka.kafka   heartbeat errored       {"namespace": "name_space", "group": "groupid", "err": "context canceled"}
logger.go:146: 2025-05-28T10:39:54.776-0400     INFO    1.kafka.kafka   assigning partitions    {"namespace": "name_space", "why": "clearing assignment at end of group management session", "how": "unassigning everything", "input": ""}
logger.go:146: 2025-05-28T10:39:54.776-0400     INFO    1.kafka.kafka   leaving group   {"namespace": "name_space", "group": "groupid", "member_id": "kgo-e4ee4cfac5d8b871e12f3743b8d22f21"}
consumer_test.go:381: got: 2 events errored, 0 processed, want: 2 errored, 26 processed
consumer_test.go:390:
        Error Trace:    /Users/rubenvanstaden/issue/apm_queue/661/apm-queue/kafka/consumer_test.go:927
[...]

PR #663, opened 2025-05-28
Update kafka and kfake

rubvs:
Closes https://github.com/elastic/apm-queue/issues/661

Do not use the same context across multiple `Produce` calls in `TestConsumerDelivery`. In doing so, the context will be cancelled after the first record is produced, dropping a records in the buffer.


rubvs:
Superseded by https://github.com/elastic/apm-queue/pull/675

Issue #640, opened 2025-02-28
Refactoring: encapsulate config via env vars into a module


inge4pres:
**Describe the current state**

The configuration layer for the Kafka queue type contains long `Commonconfig.finalize()` method, where multiple sources of configurations are mixed: file and env vars are handled together.

The code is here
https://github.com/elastic/apm-queue/blob/065e8907e47853a58615a869e74ed2923812e1bf/kafka/common.go#L204-L255

This has grown enough to justify introducing more modularity and split the responsibility of fetching configs from different sources in their own modules.
Understanding how certain configurations interact with others is difficult at the moment.

**Describe the desired state**

It looks like having a dedicated struct that handles fetching configurations from environment and dispatching properties such as TLS and SASL could help reducing the complexity of the `finalize` method.

We already have a [`saslConfigProperties`](https://github.com/elastic/apm-queue/blob/d8deb101fb600eaad0aac5d7a6485b1d4877e88b/kafka/configfile.go#L140-L144) struct to handle SASL.
It seems natural to also have a similar struct to describe TLS configs' state, and then wrap both into a new container that will expose methods to indicate the conditions that are now expressed as `if`/`else` conditions.

At a minimum, the new container would expose methods to:

* validate conflicts between envrionments variables and existing configs read from file
* provide brokers string
* provide helper to detect if the connection is in plaintext or encrypted
* provide a `*tls.Config` object to be used in encrypted setups (includes with/without `tls.InsecureSkipVerify`)

rubvs:
Move to the next iteration to have a deeper look at the franz-go upgrading issues.

PR #679, opened 2025-07-17
Refactor config structure

rubvs:
Closes https://github.com/elastic/apm-queue/issues/640

Simplify the handling of configuration options. Create a clear divide between configuration types; _environment variables_ & _config files_.

**Changes are based on the following axioms:**
- Config options set via `env` vars can be parsed completely _independent_ from those set within a file.
- The file config can be _dependent_ on env vars.
- The file config takes precedence.
- The logical interplay between env vars and file config can be handled in a separate function.

**Code structure:**
- Consists of 2 containers; `envConfig` & `fileConfig`
- Parse env vars first, since file config can be dependent on env vars.
- Merge the two containers back into `CommonConfig` by flattening them.


marclop (approved):
Looks good, thanks for the refactor

PR #709, opened 2025-09-25
ignore unknown topic warning

rubvs:
Closes https://github.com/elastic/hosted-otel-controller/issues/154

Add a config option to toggle logging the `UNKNOWN_TOPIC_OR_PARTITION` warning when monitoring consumer lag. The default will be `false`, meaning the warning will **not** be logged.

### DIFF
--- a/kafka/manager.go
+++ b/kafka/manager.go
@@ -277,6 +277,9 @@ func (m *Manager) MonitorConsumerLag(topicConsumers []apmqueue.TopicConsumer) (m
 				}
 				for partition, lag := range partitions {
 					if lag.Err != nil {
+						if lag.Err == kerr.UnknownTopicOrPartition && m.cfg.DisableUnknownTopicWarning {
+							continue
+						}
 						logger.Warn("error getting consumer group lag",
 							zap.String("group", l.Group),
 							zap.String("topic", topic),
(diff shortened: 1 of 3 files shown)


kafka/manager_test.go
```diff
@@ -771,6 +771,135 @@ func TestListTopics(t *testing.T) {
...
+		desc       string
+		logWarning bool
+	}{
+		{
+			desc:       "log UNKNOWN_TOPIC_OR_PARTITION warning for consumer lag",
+			logWarning: true,
+		},
+		{
+			desc:       "ignore UNKNOWN_TOPIC_OR_PARTITION warning for consumer lag",
+			logWarning: true,
```
isaacaflores2, line 785: Looks like we do not expect a log here. Should this be set to false?

kafka/manager.go
```diff
@@ -277,6 +277,9 @@ func (m *Manager) MonitorConsumerLag(topicConsumers []apmqueue.TopicConsumer) (m
 				}
 				for partition, lag := range partitions {
 					if lag.Err != nil {
+						if lag.Err == kerr.UnknownTopicOrPartition && !m.cfg.LogUnknownTopicWarning {
```
isaacaflores2, line 280: To avoid adding another config, what do you think of using a debug log  for the specific `UnknownTopicOrPartition` error?
```go
if lag.Err == kerr.UnknownTopicOrPartition {
	logger.Debug(...)
)
} else {
   logger.Warn(...)
}
continue
```

PR #772, opened 2026-05-26
kafka: apply topic configs via AlterTopicConfigs after creation

zmoog:
## Summary

Some Kafka-compatible systems (e.g. WarpStream) silently ignore vendor-specific configs — such as `warpstream.*` — when they are passed in the `CreateTopics` request. Standard Kafka configs like `retention.ms` work correctly at creation time, but custom configs are only reliably applied via a subsequent `IncrementalAlterConfigs` call.

### Root cause

`CreateTopics` categorises topics into three buckets and processes them separately:

- **missing** → created via `CreateTopics` API (configs passed here, but ignored by WarpStream for vendor-specific keys)
- **existing** → partition count updated via `UpdatePartitions`
- **existing** → configs updated via `AlterTopicConfigs` ✅

Only pre-existing topics went through `AlterTopicConfigs`. Newly created topics never did, so `warpstream.*` configs were never applied on first creation.

### Fix

- `createMissingTopics` now returns the list of topics that are **live** after the call (freshly created + `TopicAlreadyExists`). Topics that fail with a real error are excluded.
- `CreateTopics` merges those into the `alterTopicConfigs` call alongside the pre-existing topics, so `IncrementalAlterConfigs` is always issued for every live topic.

## Testing

- Updated `TestCreateTopics` to assert that `AlterTopicConfigs` is called for all live topics (topic0 pre-existing + topic1/3/4 from the creation response), not just the pre-existing one.
- Fixed `sortAlteredConfigs` to also sort resources by `ResourceName` for deterministic ordering.

> **Note:** This PR is a draft pending confirmation from the WarpStream/Confluent team that ignoring `warpstream.*` configs in `CreateTopics` is expected behaviour. See repro gist: https://gist.github.com/zmoog/e54f2c8e7f02e54cef5238ed04a244b2

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kafka/topiccreator.go
+++ b/kafka/topiccreator.go
@@ -133,7 +133,8 @@ func (c *TopicCreator) CreateTopics(ctx context.Context, topics ...apmqueue.Topi
 	missingTopics, updatePartitions, existingTopics := c.categorizeTopics(existing, topicNames)
 
 	var updateErrors []error
-	if err := c.createMissingTopics(ctx, span, missingTopics); err != nil {
+	createdTopics, err := c.createMissingTopics(ctx, span, missingTopics)
+	if err != nil {
 		updateErrors = append(updateErrors, err)
 	}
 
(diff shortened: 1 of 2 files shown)


zmoog:
Closing this for now as we probably won't need it. The WarpStream folks confirmed it's an issue in the CreateTopic path that they're fixing.

